PR #30, opened 2020-02-07
Cmper

jasone:
I ended up doing more work on the `Hash` module than expected, mainly because I didn't want to
implement a bunch of type-specific `hash_fold` functions using an API that was going to change soon.  It's going to be important to implement hashing before the `Map` and `Set` modules, so that we get more than 32 hash bits, but `OrdMap` and `OrdSet` come first.

The `Hash.State.seed` function can't be tested in a unit test.  We need an integration test, but I haven't figured out how to do that with OCaml/Dune yet.  Perhaps now is the time.

The `Cmper` functors aren't tested, but `OrdMap` and `OrdSet` will indirectly do so soon.

As an aside, note the `|>` operator in the `hash_fold` functions.  It makes writing pipeline code much cleaner, but it also requires that the function parameter being piped from function to function is the last parameter.  IIRC this is why the Base/Core developers were contemplating the merits of moving the `t` parameter of all functions to the end.  I'm interested in your opinion on that API design question.


jasone:
9b9977f37645425632a51824b7729b9b9e9726c4 adds integration tests for `Hash.State.seed`, as well as a resulting complete reimplementation.

jasone:
Once more, with feeling: 3a8f5f57481211ca39b30899fe866eb69fa9373d re-reimplements `Hash.State.seed` based on `getentropy()` rather than `/dev/[u]random`.

bootstrap/src/hash.c
```diff
@@ -0,0 +1,49 @@
...
+
+  caml_copy_int64(entropy);
+}
+
+// The OCaml FFI is quite low-level, such that returning (Int64.t option) is a
+// lot of trouble relative to returning an array.  As a consequence this
+// function returns an empty array to indicate error.
+//
+//                                       nbits -> [|entropy...|]
+// val hemlock_hash_state_entropy_nbits: usize -> Int64.t array
```
jasone: This may look over-engineered, but it will soon support 128-bit seeding for hashing based on [MurmurHash3](https://github.com/aappleby/smhasher) without modification.

bootstrap/src/hash.mli
```diff
@@ -3,22 +3,41 @@
 open Rudiments_int0
 
 type t
```
jasone: `Hash.t` remains opaque for the moment because 1) we're not actually reading it for anything yet, and 2) it's going to be a `u128` or the moral equivalent thereof (rather than the current `usize`), and I don't want to expose the innards until the type is right.

bootstrap/src/string.ml
```diff
@@ -85,18 +87,23 @@ module Cursor = struct
       bindex: usize;
     }
 
+    let hash_fold t state =
+      state
+      |> hash_fold t.string
+      |> Usize.hash_fold t.bindex
```
jasone: This syntax is a good example of why I changed the `hash_fold` parameter order.  Absent the `|>` operator this code is much messier and harder to follow.

src/stdlib/array.hli
```diff
@@ -17,6 +17,12 @@ module Cursor : sig
                               and type ''a $t := ''a $t
 end
 
+val hash_fold: (''a -> Hash.State.t -> Hash.State.t) -> ''a $t -> Hash.State.t
+    -> Hash.State.t
+(** [hash_fold hash_fold_a t state] incorporates the hash of [t] into [state]
+    and returns the resulting state.  Array elements are sequentially
+    hash-folded into the resulting state via [hash_fold_a]. *)
+
```
jasone: It might be a good idea to extract this into a polymorphic functor in the long run, but it's not yet clear to me what level to integrate that at.  This will be something to revisit when adding hashing for `'a list`.

bootstrap/src/hash.c
```diff
@@ -0,0 +1,49 @@
...
+  size_t nbits_in = Unsigned_long_val(a_nbits);
+  size_t nbytes_in = (nbits_in + 7) / 8;
+  size_t nuint64s = (nbytes_in + 7) / 8;
+  uint64_t entropy[nuint64s];
+  size_t nbits_out = hemlock_getentropy((uint8_t *)entropy, nbits_in);
+  uint64_t error = (nbits_out < nbits_in);
+
+  if (nbits_out == 0) {
+    const char *result = {NULL};
+    caml_alloc_array(value_of_result, &result);
```
cevans87: My c-fu has become weak, so not sure if this should be obvious. Do we need to return this?

bootstrap/src/string.ml
cevans87: Agreed. I like this way.

bootstrap/src/hash.c
jasone: Oops, regardless of whether C happens to do the right thing here, the omission of `return` was unintentional.

bootstrap/src/cmper.ml
```diff
@@ -0,0 +1,58 @@
+(* The redundancy between the .mli and .ml files cannot be extracted into an
+   interface file, apparently because the phantom witness type somehow causes
+   two incompatible-but-identical comparator types to exist. *)
+
+type ('a, 'witness) t = {
```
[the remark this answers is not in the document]
jasone: This is intended to be functionally equivalent to the comparator witnesses in [Base](https://github.com/janestreet/base).  [__Real World OCaml__](https://dev.realworldocaml.org/maps-and-hashtables.html#why-comparator-witnesses) explains the concept pretty well, better than I can.  I had to use Base maps for a while before it started feeling natural.

PR #32, opened 2020-02-08
Fix hlc compilation.

jasone:
Fix hlc.ml to fully adopt uint-to-usize renaming and related API changes.  Also
clean up the code to match its current minimal intent of verifying that an
executable can compile and execute using the bootstrap library code.

This regression was introduced by dfdaa2384213394ea3818e717c0f9e5d3eacd85d
(Rename uint/int to usize/isize, and make usize the default integer type.).

### DIFF
--- a/bootstrap/bin/hlc.ml
+++ b/bootstrap/bin/hlc.ml
@@ -1,9 +1,11 @@
 open Hemlock
 include Hemlock.Rudiments
 
-type 'a v = 'a array
-
 let _ =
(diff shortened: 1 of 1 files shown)


jasone:
Trivial CI would have caught this.  I apparently manually tested `hlc` partway through developing dfdaa2384213394ea3818e717c0f9e5d3eacd85d, but forgot to go back and do it again at the end.

PR #33, opened 2020-02-10
Add stream module

cevans87:


bootstrap/src/stream.mli
```diff
@@ -0,0 +1,89 @@
...
+val split: 'a t -> usize -> 'a t * 'a t
+(** Split the stream with elements [\[0..len)] into streams with elements
+    [\[0..n)] and [\[n..len)].  If stream contains fewer than [n] elements,
+    returns a stream with elements [\[0..len)] and and empty stream.  Split is
+    O(1), but forcing the first element of the second returned stream is
+    Ω(n)/O(n). *)
+
+val rev_split: 'a t -> usize -> 'a t * 'a t
+(** Split the stream with elements [\[0..len)] into streams with elements
+    [\[n..0)] and [\[n..len)].  If stream contains fewer than [n] elements,
```
jasone: Should this be `[(n..0\]]`?

bootstrap/src/stream.mli
```diff
@@ -0,0 +1,89 @@
...
+    returns a stream with elements [\[0..len)] and and empty stream.  Split is
```
jasone: Should this be `[(len..0\]]`?

bootstrap/src/stream.mli
```diff
@@ -0,0 +1,89 @@
...
+    O(1), but forcing the first element of either returned stream is Ω(n)/O(n).
+    *)
+
+val take: 'a t -> usize -> 'a t
+(** Return a new stream with the first [max n len] elements of the input stream.
+    *)
+
+val rev_take: 'a t -> usize -> 'a t
+(** Return a new stream with the first [max n len] elements of the input stream.
```
jasone: This should mention that the result is reversed relative to the input.

bootstrap/src/stream.mli
```diff
@@ -0,0 +1,89 @@
...
+    Forcing the first element of the returned stream is Ω(n)/O(n). *)
+
+val drop: 'a t -> usize -> 'a t
+(** Return a new stream without the first [max n len] elements of the input
+    stream.  Foricing the first element of the returned stream is Ω(n)/O(n). *)
```
jasone: s/Foricing/Forcing/

bootstrap/src/stream.ml
```diff
@@ -0,0 +1,576 @@
...
+
+let empty = lazy Nil
+
+let init n ~f =
+  let rec fn i n f = lazy begin
+    match i < n with
+    | false -> Nil
+    | true -> begin
+      let elm = f i in
+      let t = fn (Usize.succ i) n f in
```
jasone: I think `(Usize.succ i)` can be shortened to `(succ i)` here.  Only in modules where we implement `succ` do we have to be explicit about `Usize`.

bootstrap/src/stream.ml
```diff
@@ -0,0 +1,576 @@
...
+
+let hd = function
+  | lazy Nil -> halt "Empty stream has no head"
+  | lazy (Cons(elm, _)) -> elm
+
+let tl = function
+  | lazy Nil -> halt "Empty stream has no tail"
+  | lazy (Cons(_, t)) -> t
+
+let push' elm s =
```
jasone: What do you think about renaming this to `cons`?  `push'` initially made me think it was an internal function just for `push`.

Hmm, on second thought after reading through later code, I like the consistency of the `foo'` functions being non-lazy counterparts to lazy `foo` functions.  It might be worth a comment above this function to call out the convention though.

bootstrap/src/stream.ml
```diff
@@ -0,0 +1,576 @@
...
+  | (lazy Nil) -> Lazy.force t1
+  | (lazy (Cons(elm, t0'))) -> push' elm (concat t0' t1)
+end
+
+let rec take t n =
+  lazy begin
+    match t, n with
+    | _, 0 -> Nil
+    | (lazy Nil), _ -> Lazy.force t
+    | (lazy (Cons(elm, t'))), _ -> push' elm (take t' (n - 1))
```
jasone: `(n - 1)` can be written as `(pred n)`.  I don't have a strong preference on which to use, but it's probably best to be consistent about using `succ/pred` vs `+- 1`

bootstrap/src/stream.ml
```diff
@@ -0,0 +1,576 @@
...
+let rev_split t n =
+  rev_take t n, drop t n
+
+let pp pp_elm ppf t =
+  let open Format in
+  fprintf ppf "@[<h>";
+  let rec fn t = begin
+    match t with
+    | lazy Nil -> fprintf ppf "Nil"
+    | lazy(Cons(elm, t')) -> begin
```
jasone: Missing space after `lazy`.

bootstrap/src/stream.ml
```diff
@@ -0,0 +1,576 @@
...
+
+let%expect_test "init" =
+  let open Format in
+  let ppt = (pp Usize.pp) in
+  printf "@[<h>";
+  let rec test_init_up_to i n = begin
+    match i <= n with
+    | false -> ()
+    | true -> begin
+      let t = init i ~f:(fun i' -> i') in
```
jasone: `i'` would be better as `i` or `j` here.  `i'` typically indicates the value is an immediate derivative of `i`.

bootstrap/src/stream.ml
```diff
@@ -0,0 +1,576 @@
...
+  test_rev_up_to 0 5;
+  printf "@]";
+
+  [%expect{|
+    rev Nil = Nil
+    rev (0 Nil) = (0 Nil)
+    rev (0 (1 Nil)) = (1 (0 Nil))
+    rev (0 (1 (2 Nil))) = (2 (1 (0 Nil)))
+    rev (0 (1 (2 (3 Nil)))) = (3 (2 (1 (0 Nil))))
+    rev (0 (1 (2 (3 (4 Nil))))) = (4 (3 (2 (1 (0 Nil))))) |}]
```
jasone: The tests look great in general.  Note that depending on the test, there's typically no real benefit to testing beyond a size in the 0..3 range.  Reducing the input ranges reduces what we have to manually review on subsequent changes, so it's a good idea to keep the output as small as is safe.

bootstrap/src/stream.mli
```diff
@@ -0,0 +1,89 @@
...
+val init: usize -> f:(usize -> 'a) -> 'a t
+(** Initialize stream.  [init len ~f:(fun i -> ...)] lazily initializes a stream
+    of given length, where [f] provides the value for each element at given
+    index. *)
+
+(** {1 Length} *)
+
+val length: 'a t -> usize
+(** Return stream length.  This forces and counts every node and is
+ * Ω(n)/O(n). *)
```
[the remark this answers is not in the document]
cevans87: Well darn. I had a big misconception about the definition of `Θ`. I thought it was used for common-case time complexity, somewhere between best and worst case. I double checked and you're right.

bootstrap/src/stream.mli
```diff
@@ -0,0 +1,89 @@
...
+
+val concat: 'a t -> 'a t -> 'a t
+(** Concatenate two streams. *)
+
+val split: 'a t -> usize -> 'a t * 'a t
+(** Split the stream with elements [\[0..len)] into streams with elements
+    [\[0..n)] and [\[n..len)].  If stream contains fewer than [n] elements,
+    returns a stream with elements [\[0..len)] and and empty stream.  Split is
+    O(1), but forcing the first element of the second returned stream is
+    Ω(n)/O(n). *)
```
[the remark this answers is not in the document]
cevans87: Realized that the split and rev_split cases are actually `Ω(1)/O(n)` since we could have done it in a way that creates no monolith work in one of the streams. For example
```
let t = init 1_000_000_000 ~f:(fun i -> i) in
let _, t1 = split t 0 in
let _ = hd t1
```
Since we split at 0, no extra work is needed to access the first element of the second list since we didn't have to walk over any elements. This is the best-case time.

I'll just change the comment to `O(n)`.

bootstrap/src/stream.mli
cevans87: Oh, yes.

bootstrap/src/stream.ml
```diff
@@ -0,0 +1,576 @@
...
+      let t = fn (Usize.succ i) n f in
+      Cons(elm, t)
+    end
+  end in
+  fn 0 n f
+
+let length t =
+  let rec fn t i = begin
+    match t with
+    | (lazy Nil) -> i
```
cevans87: Lots of cases of unneeded parenthesis around (lazy foo). These should be removed.

PR #34, opened 2020-02-13
Add q module

cevans87:
The implementation for this module was adapted from the real-time queue
in the book "Purely Functional Data Structures" by Chris Okasaki.

### DIFF
--- a/bootstrap/src/q.ml
+++ b/bootstrap/src/q.ml
@@ -1,0 +1,199 @@
+open Rudiments
+
+type 'a t = usize * 'a Stream.t * 'a list * 'a Stream.t
+
+let empty = (0, lazy Stream.Nil, [], lazy Stream.Nil)
+
+let exec (l, f, r, s) =
+  let rec rotate (l, f, r, s) = lazy begin
+    match f, r with
+    | _, [] -> not_reached ()
+    | lazy Stream.Nil, y :: _ -> Lazy.force (Stream.push y s)
+    | lazy (Stream.Cons(elm, f')), y :: y' -> begin
+      let s' = Stream.push y s in
+      let f'' = rotate (l, f', y', s') in
+      Lazy.force (Stream.push elm f'')
+    end
+  end in
+  match s with
+  | lazy (Stream.Cons(_, s')) -> (l, f, r, s')
+  | lazy Stream.Nil -> begin
+    let f' = rotate (l, f, r, s) in
+    (l, f', [], f')
+  end
+
+let length (l, _, _, _) =
+  l
+
+let is_empty (l, _, _, _) =
+  l = 0
+
+let hd (_, f, _, _) =
+  match f with
+  | lazy Stream.Nil -> halt "Empty q has no head"
+  | lazy (Stream.Cons(elm, _)) -> elm
+
+let tl (l, f, r, s) =
+  match f with
+  | lazy Stream.Nil -> halt "Empty q has no tail"
+  | lazy (Stream.Cons(_, f')) -> exec (pred l, f', r, s)
+
+let push_back (l, f, r, s) elm =
+  exec (succ l, f, elm :: r, s)
+
+let pop t =
+  hd t, tl t
+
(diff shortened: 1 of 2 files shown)


bootstrap/src/q.mli
```diff
@@ -0,0 +1,43 @@
+(** Lazy-evaluated immutable q.
```
jasone: s/q/queue/ here.  The rest of the docs look fine though wrt q vs queue.

bootstrap/src/q.ml
```diff
@@ -0,0 +1,202 @@
...
+let push_back (l, f, r, s) elm =
+  exec (succ l, f, elm :: r, s)
+
+let pop t =
+  hd t, tl t
+
+let pp pp_elm ppf (l, f, r, s) =
+  let open Format in
+  fprintf ppf "@[<h>";
+  fprintf ppf "(len=%a, f=%a, r=%a, s=%a)"
```
jasone: The spaces in the format specifier should have `@` prefixes so that the Format machinery knows they are acceptable places to insert line breaks.  This isn't necessary in test code, but we should try to keep the `pp` functions capable of automatic wrapping for use in non-trivial formatted output cases.

  `"(len=%a,@ f=%a,@ r=%a,@ s=%a)"`

bootstrap/src/q.ml
```diff
@@ -0,0 +1,202 @@
+open Rudiments
+
+type 'a t = usize * 'a Stream.t * 'a list * 'a Stream.t
+
+let empty = (0, lazy Stream.Nil, [], lazy Stream.Nil)
+
+let exec (l, f, r, s) =
+  let rec rotate (l, f, r, s) = lazy begin
+    match f, r with
+    | _, [] -> assert false
```
jasone: We have `not_reached ()` for such uses.

bootstrap/src/q.ml
```diff
@@ -0,0 +1,202 @@
...
+    end
+  end in
+  match s with
+  | lazy (Stream.Cons(_, s')) -> (l, f, r, s')
+  | lazy Stream.Nil -> begin
+    let f' = rotate (l, f, r, s) in
+    (l, f', [], f')
+  end
+
+let length (l, _, _, _) = l
```
jasone: Thus far my preferred style for top-level declarations like this has been to put the function body on a separate line even if it's simple.  That way we can tell at a glance whether the body is a constant vs a function.  So the `empty` declaration above is fine, but this one needs a line break.  Does that seem like a worthwhile distinction?  I could be convinced either way.

bootstrap/src/q.ml
```diff
@@ -0,0 +1,202 @@
...
+  printf "@]";
+
+  [%expect{| empty = (len=0, f=Nil, r=[], s=Nil) |}]
+
+let%expect_test "length" =
+  let open Format in
+  let ppt = (pp Usize.pp) in
+  printf "@[<h>";
+  let rec test_length_up_to i n t = begin
+    match i < n with
```
jasone: I'd expect `*_up_to` to apply to an open bound (`match i <= n with`), in order to be consistent with the semantics of for loop bounds `to` and `downto`.  TBH I'm not crazy about the keyword names, but the open-open semantics are actually important, e.g. in the `downto` case, since otherwise there would be no way to express "down to 0 inclusive" on unsigned integers.

cevans87: Do you mean the naming like "test_length_up_to"? I'd be fine with just changing all those to fn.

bootstrap/src/q.ml
cevans87: Seems like a decent convention.

bootstrap/src/q.ml
jasone: I'm fine with just `fn` here, or a change to the names (e.g. `test_length` here).

cevans87: Cool. Some of these comments apply to the stream module that I've already commited. I've fixed these in there as well and will make a final "Use consistent idioms" PR when we're done with deq.

PR #37, opened 2020-02-15
Add deq module

### LINKED ISSUE #25: Implement Deq module
Implement persistent real-time double-ended queues, using the techniques described in __Purely Functional Data Structures__, by Chris Okasaki.  The `Deq` API should have substantial overlap with the `List` API, with at least the following additions:

- `val push_back: 'a t -> 'a -> 'a t`: Push element onto the back and return the updated dequeue.
- `val back: 'a t -> 'a`: Return the back element.
- `val pop_back: 'a t -> 'a * 'a t`: Pop the back element and return it, along with the updated dequeue.

cevans87:
Close #25

The implementation for this module was adapted from the real-time
double-ended queue in the book "Purely Functional Data Structures" by
Chris Okasaki.

### DIFF
--- a/bootstrap/src/deq.ml
+++ b/bootstrap/src/deq.ml
@@ -1,0 +1,322 @@
+open Rudiments
+
+type 'a t =
+  usize * 'a Stream.t * 'a Stream.t * usize * 'a Stream.t * 'a Stream.t
+
+(* Increment size for rebuilding; must be either 2 or 3. Reference: Double-ended
+   queues section in Purely Functional Data Structures by Chris Okasaki. *)
+let c = 2
+
+let empty = (
+  0, lazy Stream.Nil, lazy Stream.Nil,
+  0, lazy Stream.Nil, lazy Stream.Nil
+)
+
+let length (lf, _, _, lr, _, _) =
+  lf + lr
+
+let is_empty (lf, _, _, lr, _, _) =
+  lf + lr = 0
+
+let exec1 = function
+  | lazy (Stream.Cons(_, s')) -> s'
+  | s -> s
+
+let exec2 s =
+  exec1 (exec1 s)
+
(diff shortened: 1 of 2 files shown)


bootstrap/src/deq.ml
```diff
@@ -0,0 +1,320 @@
+open Rudiments
+
+type 'a t =
+  usize * 'a Stream.t * 'a Stream.t * usize * 'a Stream.t * 'a Stream.t
+
+let c = 2
```
jasone: A brief comment here would be useful to make it clear this isn't just a sloppy magic number.  Something like:

```ocaml
(* Increment size for rebuilding; must be either 2 or 3. *)
```

bootstrap/src/deq.ml
```diff
@@ -0,0 +1,320 @@
...
+    | lazy Stream.Nil -> ()
+    | lazy (Cons(elm, lazy Stream.Nil)) -> fprintf ppf "%a" pp_elm elm;
+    | lazy (Cons(elm, s')) -> begin
+      fprintf ppf "%a,@ " pp_elm elm;
+      fn s'
+    end
+  end in
+  fprintf ppf "(";
+  fn f;
+  fprintf ppf "@ |@ ";
```
jasone: The `( ... | ... )` format surprised me out at first, but I really like it.

bootstrap/src/deq.mli
```diff
@@ -0,0 +1,59 @@
+(** Lazy-evaluated immutable double-ended queue.
+
+     The deq is an immutable double-ended queue container. All operations on deq
+     are amortized via lazy execution and are O(1).
```
jasone: It looks like this might be over-indented.

cevans87:
I moved "Close #25" to the bottom. It shows in the new commit message but not at the top in this discussion.

PR #38, opened 2020-02-18
Fix warnings for macOS build.

jasone:
getentropy() on macOS requires the sys/random.h header; include it.

Add missing 'return' keyword.

### DIFF
--- a/bootstrap/src/hash.c
+++ b/bootstrap/src/hash.c
@@ -1,6 +1,10 @@
 #include <stdbool.h>
 #include <stddef.h>
 #include <unistd.h>
+#ifdef __APPLE__
+// For getentropy().
+#  include <sys/random.h>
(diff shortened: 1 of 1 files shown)


cevans87:
It'd be good for these warnings to show up in our CI, right? Do you recommend any changes there?

jasone:
The warnings did show up in the detailed logs, but they didn't cause the build to fail.  My understanding was that Dune ordinarily fails on any warning (see e.g. https://dune.readthedocs.io/en/latest/faq.html#how-to-make-warnings-non-fatal), but that seems to only apply to OCaml code.  It may be possible to manually specify the C compiler command line, but I'm not sure it's worth going to that amount of trouble given how little C code we're likely to integrate into the bootstrap compiler.

PR #40, opened 2020-02-19
Implement MurmurHash3-based hashing.

### LINKED ISSUE #8: Implement Hash module
A preliminary `Hash` module is already in place, but it relies on OCaml's polymorphic built-in hashing functions.  Enhance the implementation to directly support built-in basic types.

### LINKED ISSUE #39: Add bit_ssr to unsigned integer modules.
`bit_ssr` is a low-level operation that is well defined regardless of the integer type signedness.  There are legitimate uses for `bit_ssr` on unsigned integers, so we should support it.

jasone:
Implement hashing based on the 128-bit variant of MurmurHash3.  This requires
the addition of u128 for API purposes.  u128 in turn builds on top of u64, i64,
and u32.

Add bit_ssr to unsigned integer types, because behavior is well defined, and the
function has legitimate uses.

Expand testing of integer types and fix various latent bugs.
- Floating point conversion was badly broken in many cases.  Improve it enough
  to allow correct implementation for u128, but don't bother covering up OCaml's
  unfortunate behavior on overflow for intnb-based integer types.
- Fix edge condition handling for {floor,ceil}_{pow2,lg}.

Resolve #8.
Resolve #39.


jasone:
A huge portion of this diff is nearly identical unit tests for all the integer types.  In the past I left out a bunch of tests because of how much redundancy they require, but with the addition of `u128`, `u64`, and `i64` I was no longer confident in test-free code.  You can probably review tests for just one of the types without missing anything significant.

bootstrap/src/intnb.ml
```diff
@@ -1,12 +1,72 @@
+open Rudiments_functions
 open Intnb_intf
 
+module Make_derived (T : I_derived) : S_derived with type t := T.t = struct
```
jasone: I debated whether to extract this code into a functor, but after repeatedly fixing issues in four nearly identical copies (and sometimes missing one or more copies) I decided to go for it.  The `**` operator could also be extracted into this functor, but doing so would require the functor to additionally know whether it is used in a signed context.  Thoughts on that one?

bootstrap/src/u128.ml
```diff
@@ -0,0 +1,1281 @@
...
+    let x = bit_or x (bit_usr x 8) in
+    let x = bit_or x (bit_usr x 16) in
+    let x = bit_or x (bit_usr x 32) in
+    let x = bit_or x (bit_usr x 64) in
+    bit_pop (bit_not x)
+
+  let bit_ctz t =
+    bit_pop (bit_and (bit_not t) (t - one))
+
+  let div_mod t0 t1 =
```
jasone: This is the function that motivated adding `i64` and `u32`.  I initially got it working without these types, but only after much gnashing of teeth due to subtleties regarding truncation and sign extension.

bootstrap/src/u128.ml
```diff
@@ -0,0 +1,1281 @@
+include Rudiments_int
```
jasone: `u128` was a significant development effort, to the point that it was tempting to just make the `Hash` API use `(u64 * u64)` instead of `u128`.  However, we're probably also going to need `u256` to support cryptographic hashing (SHA-256 or SHA-3-256) for uniquely identifying source code, API signatures, etc., so it felt worthwhile to do a full implementation.  `u256` will probably require extracting various `u128` code for sharing, but I'd rather not try to anticipate the particulars right now.

bootstrap/src/u128.ml
```diff
@@ -0,0 +1,1281 @@
...
+      | true -> ()
+      | false -> begin
+          fn_i 0 j U64.zero;
+          fn_j (Rudiments_int.succ j)
+        end
+    end in
+    fn_j 0;
+    of_arr product
+
+  let of_string s =
```
jasone: There was nothing in OCaml to piggyback this on, so here we have a state machine for lexing `u128`.  We'll need something similar for the compiler, at which point we can hopefully unify the implementations.

bootstrap/src/hash.mli
```diff
@@ -27,17 +28,41 @@ module State : sig
...
+  module Gen : sig
+    type outer = t
 
-  val hash_fold_float: float -> t -> t
-  (** [hash_fold_float f state] incorporates the hash of [f] into [state] and
-      returns the resulting state. *)
+    type t
+    (** Hash state generator type. *)
+
+    val init: outer -> t
```
jasone: It occurred to me while thinking about pseudo-random number generation that `Hash.State.seed` isn't really specific to hashing; we could use that same entropy for PRNG seeding.  What do you think about moving `Hash.state.seed` to, say, `Entropy.seed`, and changing `Hash.State.Gen.init` to take a `u128` rather than `Hash.State.t` parameter?

jasone: I went ahead and did something close to this.  We now have `Entropy.seed` which is explicitly `u128`, and `Hash.State.seed` remains as an opaque state.  `HEMLOCK_ENTROPY` is now interpreted as a `u128` constant, which has the major benefit that if an application logs the value of `Entropy.seed`, it's possible to reproduce the execution by feeding in the same seed.  Before this change we were string-hashing the contents of `HEMLOCK_ENTROPY`, so even if the application logged the seed, there was no way to feed it back in to a subsequent run.

jasone:
Committing, per offline discussion.

Issue #27, opened 2020-02-04
Run code through ocp-indent.


jasone:
Lots of code has somehow slipped in without proper indentation.  Standardize formatting by running all the code through ocp-indent and manually adjust things like line breaks as necessary to improve overall code consistency.

cevans87:
I also had the thought of creating a github action that shows scary-red and possibly blocks merging PRs so that we're always properly indented.

jasone:
A GitHub action for this would be great.  I'd also like to prohibit trailing whitespace if `ocp-indent` doesn't automatically strip it out.

In a couple places we'll need to change style to make automatic formatting work nicely.  The one I'm currently aware of is

```ocaml
(*******************************************************************************
 * Begin tests.
 *)
```
The last line's leading whitespace is stripped by `ocp-indent`.

This indents as expected:

```ocaml
(******************************************************************************)
(* Begin tests. *)
```

jasone:
I suggest we tackle this as soon as we have #8, #25, and #39 integrated.

cevans87:
Another thought I have is that we might want two separate actions for ocp-indent. One for pull_requests and one that's periodically run on master or based on new code being committed to master.

I think it'd be best if a pull_request ocp-indent action just catches funny problems in the lines added or modified by the current PR. If there happen to be indentation problems already in master, this action wouldn't block us from committing new, clean code.

For the periodic/on_merge to master action, I think ocp-indent could just run on everything and show a status that we do or don't have indentation problems in master.

jasone:
Actually, #8 and #39 aren't blockers for this issue, so go ahead.

jasone:
I think our happy steady state is code that `ocp-indent` makes no changes to.  But as you say, if you write the action to only complain about formatting issues in changes, then we can make the transition incrementally with minimal friction even if other changes are intermingled.  Presumably once we do a full manual pass and commit the formatting changes, it will never again be necessary to consider formatting issues outside of changes.

PR #43, opened 2020-02-21
Add ocp-indent checker for pull-requests.

cevans87:
This script checks that lines modified in a PR (compared to
origin/master) are indented as suggested by ocp-indent. If ocp-indent
would have done anything different with modified lines in this PR, the
check fails and prints suggested changes from ocp-indent.

ref #27

### DIFF
--- a/.github/scripts/check_ocp_indent.py
+++ b/.github/scripts/check_ocp_indent.py
@@ -1,0 +1,147 @@
+#!/usr/bin/env python3
+
+from __future__ import annotations
+from asyncio import gather, run
+from asyncio.subprocess import create_subprocess_exec, PIPE, STDOUT
+from dataclasses import dataclass
+import logging
+from pathlib import Path
+import re
+import sys
+from typing import FrozenSet, List, Optional, Pattern, Text, Tuple
+
+
+log = logging.getLogger(__name__)
+stream_handler = logging.StreamHandler(sys.stdout)
+stream_handler.setLevel(logging.INFO)
+log.setLevel(logging.INFO)
+log.addHandler(stream_handler)
+
+
+async def get_text(
+        cmd: Text,
+        stdin: Optional[Text] = None,
+        error_codes: FrozenSet[int] = frozenset({0})
+) -> Text:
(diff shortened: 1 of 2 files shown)


cevans87:
Here's roughly what the checker looks like when it fails.
https://github.com/cevans87/Hemlock/runs/460833798?check_suite_focus=true

bootstrap/src/deq.ml
```diff
@@ -9,7 +9,7 @@ let c = 2
 
 let empty = (
   0, lazy Stream.Nil, lazy Stream.Nil,
-  0, lazy Stream.Nil, lazy Stream.Nil
+  1, lazy Stream.Nil, lazy Stream.Nil
```
jasone: This looks to be unintentionally included in the pull request.  Resubmit without this change?  I'll review the rest of the code later this afternoon.

.github/scripts/check_ocp_indent.py
```diff
@@ -0,0 +1,147 @@
...
+        assert 'remotes/origin/master' in lines, 'local checkout of origin/master is required'
+
+        # Names of files changed since master.
+        lines = await get_lines('git diff --name-only remotes/origin/master')
+
+        # Absolute paths of all OCaml files changed since master.
+        paths = tuple(
+            path for path in [repo_path / Path(line) for line in lines]
+            if path.suffix in {'.ml', '.mli'}
+        )
```
jasone: Wow, Python 3 is a very different beast than Python 2.  It's really cool how the generator expression works in conjunction with async/await.

.github/scripts/check_ocp_indent.py
```diff
@@ -0,0 +1,147 @@
+#!/usr/bin/env python3.8
```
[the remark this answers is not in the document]
cevans87: Ideally I'd have a setup.cfg and setup.py to indicate that the version should be >= 3.8. I omitted those because I was being lazy and because I didn't need to specify any external dependencies. However, the way I did this here may leave someone confused in the future if they try to run against <3.8 and see runtime exceptions because of some of the features I used here.

I'll add a minimal setup.py before committing this to do it right.

PR #53, opened 2020-03-11
Avl

jasone:
The core is solid, but various ancillary functions still need tests.

### DIFF
--- a/bootstrap/src/basis/array.ml
+++ b/bootstrap/src/basis/array.ml
@@ -100,6 +100,11 @@ module Seq = struct
     type 'a elm
     val to_array: 'a t -> 'a elm outer
   end
+  module type S_poly2 = sig
+    type ('a, 'b) t
+    type 'a elm
+    val to_array: ('a, 'b) t -> 'a elm outer
+  end
 
   module Make_mono (T : Seq_intf.I_mono_def) : S_mono with type t := T.t
                                                        and type elm := T.elm =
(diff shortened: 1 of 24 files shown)


jasone:
I just did a quick review, and the only issue I found is that `Ordset.filter[i]` need tests.

jasone:
This is ready to go.  Regarding meaningful review that doesn't suck up huge amounts of time:

- The `i2` type was something I added prior to switching from imbalance-based AVL tree logic to height-based AVL tree logic.  It's an awful choice for the first signed narrowed (<63-bit) integer type to add, but it's what I needed at the time, and I debugged it pretty thoroughly.  As such, I think it's worth including either `i2` or something similar like `i32` to make sure narrow signed integers continue to work.  I don't have strong opinions on what to do here, so feedback welcome.
- The poly2 commits are all about extracting common code into modules to avoid code duplication.  They were very painful to figure out how to to write, but they compile now and they have good test coverage.  Read at your leisure or skip them.
- The Ordset commit is the meat of this pull request.  The code is pretty thoroughly tested, and I'm very pleased with the overall code coherency, but since I'm still close to the code I can't tell what is non-obvious.  I'm fine with documenting the hell out of the code for the sake of maintainability, so please call out things that aren't apparent.

jasone:
Restructured `Ordset.t` to differentiate between `Node` for internal nodes and `Leaf` for leaves.  This has almost no impact on code complexity, but it saves a lot of memory.

PR #55, opened 2020-03-20
Map

jasone:


bootstrap/src/basis/map_intf.ml
```diff
@@ -0,0 +1,402 @@
...
+  val drop: 'k -> ('k, 'v, 'cmp) t -> ('k, 'v, 'cmp) t
+  (** [drop k t] returns a map that is equivalent to the difference of [t]
+      relative to the singleton map containing key [k].  O(lg n) time complexity
+      if ordered, O(1) time complexity if unordered. *)
+
+  val unbind: 'k -> ('k, 'v, 'cmp) t -> ('k, 'v, 'cmp) t
+  (** [unbind k t] returns a map that is equivalent to the difference of [t]
+      relative to the singleton map containing key [k], where [k] is a key in
+      [t].  Halts if [k] is not a key in [t].  O(lg n) time complexity if
+      ordered, O(1) time complexity if unordered. *)
```
cevans87: As you mentioned before, the obvious pairing of names like insert/insert_hlt, replace/replace_htl and remove/remove_hlt had a nice simplicity to it, but it was unfortunate that the usual go-to function would be "replace" instead of "insert". I'm just going to put alternative ideas down for the pairs of functions. If we can find the right words, I'd be happier with foo/foo_hlt named pairs to keep the behaviour more obvious.
- augment/bind (formerly insert/insert_hlt)
  - offer/offer_hlt (less forceful than some names for in more-forceful "set/rebind" alternatives)
  - fill/fill_hlt (evokes the idea of filling a gap, or halting if it can't)
  - assign/assign_hlt
  - insert/insert_hlt (really wasn't bad. reinsert/reinsert_hlt might go better with this than replace/replace_hlt did)
- set/rebind
  - cover/cover_hlt (like covering up the old key)
  - override/override_hlt (I think this one is bad. The word is used for very different things in other languages)
  - reassign/reassign_hlt
  - reinsert/reinsert_hlt
- drop/unbind
  - drop/drop_hlt
  - unassign/unassign_hlt

jasone: remove/remove_hlt isn't too bad for the drop/unbind case.  I'd be happy with the following, since it's perfectly consistent with the Set API:
- insert/insert_hlt
- xxx/xxx_hlt
- remove/remove_hlt
What could xxx be, such that it's obviously the right choice for typical insertions?

cevans87: My favorite for about a the last whole minute (as someone who's used a db) is upsert/upsert_hlt

cevans87: Though upsert/upsert_hlt might be misleading. Sadly, upsert/update_hlt would be less of a lie.

jasone: This is the first time I've encountered 'upsert'.  Its presence in the API reflects strongly on the limits of 'insert', which is great.

cevans87: update/update_hlt would look less alien than upsert/upsert_hlt, and I think people would have no trouble understanding what happens on a conflict. Still, I think people might accidentally expect insert/insert_hlt to do the same thing.

cevans87: Another way to lead people away from the augment/bind calls is to make them obviously weaker. Like weak_insert/weak_insert_hlt.

jasone: My favorite at the moment:
- insert/insert_hlt
- upsert/update_hlt
- remove/remove_hlt

cevans87: The more I look at it, I realize that the current set/rebind isn't as perfect a pair as the other two pairs anyway.
- insert (fails silently)
- insert_hlt (fails with halt)
- upsert (never fails. Always succeeds at modifying the map)
- update_hlt (fails with halt)
- remove (fails silently)
- remove_hlt (fails with halt)

cevans87: > My favorite at the moment:
> 
>     * insert/insert_hlt
> 
>     * upsert/update_hlt
> 
>     * remove/remove_hlt

Currently my favorite as well.

PR #56, opened 2020-03-25
Add {Map,Ordmap}.update .

jasone:


bootstrap/src/basis/ordmap.ml
```diff
@@ -1214,24 +1220,27 @@ let%expect_test "mem,get,insert,insert_hlt" =
...
     | [] -> ()
     | k :: ks' -> begin
         assert (not (mem k ordmap));
         assert (Option.is_none (get k ordmap));
         let v = k * 100 in
-        let ordmap' = upsert ~k ~v ordmap in
-        assert (mem k ordmap');
-        assert ((get_hlt k ordmap') = v);
+        let ordmap' = update ~k ~v ordmap in
+        assert (not (mem k ordmap'));
```
cevans87: Does it matter that we're not also testing the successful `update` call? Looks like we're only testing the silent fail behavior.

jasone: Good observation.  I'll add testing of the other path too.

PR #57, opened 2020-03-26
of_list_rev

jasone:
This function was apparently never used before.

### DIFF
--- a/bootstrap/src/basis/array.ml
+++ b/bootstrap/src/basis/array.ml
@@ -193,10 +193,9 @@ module Seq = struct
     with type 'a t := 'a T.t
      and type 'a elm := 'a T.elm = struct
     let to_array t =
-      let l = T.length t in
-      match l with
+      match T.length t with
       | 0 -> [||]
-      | _ -> begin
+      | l -> begin
           let rec fn t a i = begin
             let elm, t' = T.next t in
             let () = Stdlib.Array.set a i elm in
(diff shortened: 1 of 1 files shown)


bootstrap/src/basis/array.ml
```diff
@@ -121,10 +121,9 @@ module Seq = struct
                                                        and type elm := T.elm =
   struct
     let to_array t =
-      let l = T.length t in
-      match l with
+      match T.length t with
       | 0 -> [||]
-      | _ -> begin
+      | l -> begin
```
cevans87: I scratched my head for a while wondering how `0` and `1` were the only cases. Finally realized this is `l` as in lookalike.

jasone: Here's a nickel; get yourself a better font. :wink:

cevans87: It's the github default!

bootstrap/src/basis/array.ml
```diff
@@ -209,18 +205,19 @@ module Seq = struct
           end in
           let elm, t' = T.next t in
           let a = Stdlib.Array.make l elm in
```
cevans87: Kinda weird that `Stdlib.Array.make` initializes every single element to `elm` before we make a second pass and correct it all. Would it make more sense to use `Stdlib.Array.init`?

jasone: Yeah, this kinda sucks.  The problem we have here is that `Seq` is a pure external iterator, and the only way we can use it with `Stdlib.Array.init` is to lexically bind a mutable variable.  I have a different mechanism in mind for the Hemlock version (https://github.com/BranchTaken/Hemlock/blob/master/src/basis/array.hl#L137)

jasone: Incidentally, the whole reason we pervasively use `Seq` is to avoid double initialization, across a variety of data types.  Absent this rough spot in the bootstrap library, I think `Seq` is turning out to serve that purpose well.

PR #60, opened 2020-03-28
Implement {Ordmap,Map,Ordset,Set}.{subset,disjoint} .

jasone:
These functions aren't overly important for maps, but they should definitely be provided for sets.

### DIFF
--- a/bootstrap/src/basis/map_intf.ml
+++ b/bootstrap/src/basis/map_intf.ml
@@ -133,9 +133,19 @@ module type S = sig
 
   val equal: ('v -> 'v -> bool) -> ('k, 'v, 'cmp) t -> ('k, 'v, 'cmp) t -> bool
   (** [equal veq t0 t1] returns [true] if [t0] and [t1] contain identical key
-      sets and identical key-value mappings as determined by [t]'s key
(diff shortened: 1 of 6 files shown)


bootstrap/src/basis/map_intf.ml
```diff
@@ -137,6 +137,17 @@ module type S = sig
       comparator and the [veq] value comparison function, [false] otherwise.
       O(n) time complexity. *)
 
+  val subset: ('v -> 'v -> bool) -> of_:(('k, 'v, 'cmp) t) -> ('k, 'v, 'cmp) t
```
jasone: I'm wondering if the `~of_` named parameter is a design mistake.  Consider that `diff`'s parameter order matters, but there we don't name either parameter.  It seems to me the same design considerations hold here.

jasone:
Committing this, since I have dependent code in another repo.

bootstrap/src/basis/map_intf.ml
cevans87: I was reading the `of_` version this morning and actually had to think about it a bit. The `t0 t1` was immediately a bit more obvious. Probably because most of the api looks more like the latter.

PR #61, opened 2020-03-30
Add choose[_hlt] to maps and sets.

jasone:
I didn't plan on adding these, but discovered while writing 'remove' benchmarks that there was no straightforward way to choose an arbitrary element of an unordered  map/set.  For ordered maps/sets we have `nth`, but the best we can do for unordered maps/sets absent `choose` is to `fold_until` and terminate after the first iteration.

### DIFF
--- a/bootstrap/src/basis/map.ml
+++ b/bootstrap/src/basis/map.ml
@@ -190,6 +190,20 @@ let get_hlt a t =
 let mem a t =
   Option.is_some (get a t)
 
+let choose t =
+  let rec fn node = begin
+    match node.elms_kv, node.elms_child with
+    | [||], [||] -> None
+    | [||], _ -> fn (Array.get 0 node.elms_child)
+    | _, _ -> Some (Array.get 0 node.elms_kv)
+  end in
+  fn t.root
+
+let choose_hlt t =
+  match choose t with
+  | Some (k, v) -> k, v
+  | None -> halt "Empty map"
+
 let insert_impl k ~f t =
   let open Cmper in
   (* Create a collision node, with keys canonically ordered (sorted). *)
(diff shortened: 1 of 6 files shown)


bootstrap/src/basis/map.ml
```diff
@@ -1552,6 +1566,31 @@ let%expect_test "of_alist,to_alist" =
...
+        assert ((length map') = (length map'') + 1);
+        map''
+      end
+  end in
+  let e = empty (module UsizeTestCmper) in
+  let _ = test 100 0 e in
+  printf "@]";
+
+  [%expect{|
+    |}]
```
cevans87: Guessing it doesn't make sense to print and check for the result of `choose_hlt` since the result of `choose` changes based on the hash seed? Wondering what we can do here to make it glaringly obvious the test really did something.

bootstrap/src/basis/ordmap.ml
```diff
@@ -1671,6 +1689,31 @@ let%expect_test "of_alist,to_alist,to_array" =
...
+        assert ((length ordmap') = (length ordmap'') + 1);
+        ordmap''
+      end
+  end in
+  let e = empty (module Usize) in
+  let _ = test 100 0 e in
+  printf "@]";
+
+  [%expect{|
+    |}]
```
cevans87: Unlike the `Map` and `Set` tests, this one does happen deterministically. Right? Would it be valuable to print stuff as we go?

jasone: I sat on this pull request for a couple days, contemplating whether to add explicit output to this test.  My current thought is that it isn't necessary.  `choose` only promises to return an arbitrary binding, and this test is good enough to demonstrate that this happens.  The current implementation actually provides the stronger guarantee that the arbitrary choice is stable for equal maps (which is part of OCaml's documented semantics), but I'm not convinced that's an important guarantee.  We may want to return the root in the future, and this test will still work correctly with that change.

PR #62, opened 2020-03-31
No rotate

jasone:
Currently based on the `choose` diff.


jasone:
Hold off on this one for now.  I've found some additional optimization opportunities that will change this diff quite a bit.

PR #65, opened 2020-04-01
Add native implementations of U64.bit_{pop,clz,ctz}.

jasone:


bootstrap/src/basis/u64.c
```diff
@@ -0,0 +1,20 @@
...
+CAMLprim value
+hemlock_u64_bit_clz(value a_x) {
+  uint64_t x = Int64_val(a_x);
+  return Val_long((x == 0) ? 64 : __builtin_clzl(x));
+}
+
+CAMLprim value
+hemlock_u64_bit_ctz(value a_x) {
+  uint64_t x = Int64_val(a_x);
+  return Val_long((x == 0) ? 64 : __builtin_ctzl(x));
```
cevans87, line 19: Why does the check for zero need to be here? Guessing this speeds up one of our existing use cases?

jasone: The result of `__builtin_ctzl()` is undefined for the `(x == 0)` case.  I didn't initially notice it in the documentation (it _is_ documented), but the unit tests caught the bug.

PR #66, opened 2020-04-01
Pathless fold

jasone:


jasone:
Big diff, but only `Ordmap`'s `fold_until` and `fold_right_until` functions are non-trivial new code.

bootstrap/src/basis/map_intf.ml
```diff
@@ -413,6 +410,18 @@ module type S_ord = sig
...
+  val min_elm: cmp:(('k * 'v) -> ('k * 'v) -> Cmp.t) -> ('k, 'v, 'cmp) t
+    -> ('k * 'v) option
+  (** [min_elm ~f t] iterates from left to right over [t] and returns [Some
+      (k, v)] for the first element which always compares as [Cmp.Lt] or
+      [Cmp.Eq], or [None] if [t] is empty. *)
+
+  val max_elm: cmp:(('k * 'v) -> ('k * 'v) -> Cmp.t) -> ('k, 'v, 'cmp) t
+    -> ('k * 'v) option
+  (** [max_elm ~f t] iterates from left to right over [t] and returns [Some
+      (k, v)] for the first element which compares as [Cmp.Eq] or [Cmp.Gt], or
```
cevans87, line 422:
```suggestion
      (k, v)] for the first element which always compares as [Cmp.Eq] or [Cmp.Gt], or
```
Just to keep it consistent with `min_elm`. Improves meaning. Guessing this suggestion exceeds line width.

cevans87: Alternatively
```first element which always compares as ...``` to ```first element with compares as [Cmp.Eq] or [Cmp.Gt] to all other elements```
No strong preference here. Just might not be super clear what the element is actually being compared with.

File: bootstrap/src/basis/ordmap.ml [hunk 2]
```diff
@@ -324,9 +324,137 @@ module T = struct
...
+  let rec fn accum f = function
+    | Empty -> accum, false
+    | Leaf {k; v} -> f accum (k, v)
+    | Node {l; k; v; n=_; h=_; r} -> begin
+        match fn accum f l with
+        | accum', true -> accum', true
+        | accum', false -> begin
+            match f accum' (k, v) with
+            | accum'', true -> accum'', true
+            | accum'', false -> fn accum'' f r
```
cevans87, line 339:
```suggestion
        let accum' stop = fn accum f l in
        match stop with
        | true -> accum', true
        | false -> begin
            let accum'' stop' = f accum' (k, v) in
            match stop' with
            | true -> accum'', true
            | false -> fn accum'' f r
```
Does it matter that we're matching on something that doesn't change between the match branches? The way it is now looks fine to me, just wonder if it's better to pull the `accum` out when we're not actually matching on it.

bootstrap/src/basis/ordmap.ml
```diff
@@ -324,9 +324,137 @@ module T = struct
...
+  let rec fn accum f = function
+    | Empty -> accum, false
+    | Leaf {k; v} -> f (k, v) accum
+    | Node {l; k; v; n=_; h=_; r} -> begin
+        match fn accum f r with
+        | accum', true -> accum', true
+        | accum', false -> begin
+            match f (k, v) accum' with
+            | accum'', true -> accum'', true
+            | accum'', false -> fn accum'' f l
```
cevans87, line 356: Same as above.

bootstrap/src/basis/set_intf.ml
```diff
@@ -305,6 +302,16 @@ module type S_ord = sig
...
 
+  val min_elm: cmp:('a -> 'a -> Cmp.t) -> ('a, 'cmp) t -> 'a option
+  (** [min_elm ~f t] iterates from left to right over [t] and returns [Some a]
+      for the first element which always compares as [Cmp.Lt] or [Cmp.Eq], or
+      [None] if [t] is empty. *)
+
+  val max_elm: cmp:('a -> 'a -> Cmp.t) -> ('a, 'cmp) t -> 'a option
+  (** [max_elm ~f t] iterates from left to right over [t] and returns [Some a]
+      for the first element which compares as [Cmp.Eq] or [Cmp.Gt], or [None] if
+      [t] is empty. *)
```
cevans87, line 313: Same as in map_intf.ml

bootstrap/src/basis/ordmap.ml
```diff
@@ -324,9 +324,137 @@ module T = struct
...
+
+let findi_map ~f t =
+  foldi_until t ~init:None ~f:(fun i _ (k, v) ->
+    match f i (k, v) with
+    | None -> None, false
+    | Some a -> Some a, true
+  )
+
+let min_elm ~cmp t =
+  fold t ~init:None ~f:(fun accum (k, v) ->
```
cevans87, line 435: I don't quite understand why we have to do a full fold here (same for `max_elm`). If this is an ordmap, can't you just take the leftmost elm here? Basically, a `fold_until` that returns false on the first elm?

jasone: Looks like the docs are confusing.  The intent is to return the leftmost element in the set of elements which would return `Lt` or `Eq` with respect to all other elements.  In other words, a full pass is required, but if multiple minima exist, the leftmost one is returned rather than an arbitrary one.

bootstrap/src/basis/map_intf.ml
jasone: How about `leftmost element which compares as [Cmp.Eq] or [Cmp.Gt] to all other elements`?

bootstrap/src/basis/ordmap.ml
cevans87: Oh, my mistake. I mixed this up with key comparison, no value comparison.

bootstrap/src/basis/map_intf.ml
jasone: I changed all instances of this documentation to `leftmost element which always compares as [Cmp.Eq] or [Cmp.Gt]`.

jasone:
Feedback incorporated.  There was another instance of the `min_elm`/`max_elm` documentation in `Container_common_intf` that I updated as well.

PR #70, opened 2020-04-04
Untuple

jasone:
Probably not for commit.  #69 removes about half of the unnecessary tuple creation in the `Ordmap` `insert`/`remove` code paths; this removes most of the remaining unnecessary tuple creation.  I'm currently inclined to leave this change out, and plan on optimizing away such boxing/unboxing in the Hemlock compiler.  Thoughts?


cevans87:
Agreed that it might be better to leave it alone for now. It shouldn't be a problem in Hemlock, so might as well not make the code uglier to side-step OCaml oddities. We may come back and do this again when we want to create a serious benchmark against Hemlock.

PR #71, opened 2020-04-14
Add File module

cevans87:
ref #52

### DIFF
--- a/bootstrap/src/basis/array.ml
+++ b/bootstrap/src/basis/array.ml
@@ -392,6 +392,52 @@ let of_list_rev ?length list =
   in
   Array_of_list_rev.to_array (Array_of_list_rev.init length list)
 
+module Array_of_stream_common = struct
+  type 'a t = {
+    stream: 'a Stream.t;
+    length: usize;
+  }
+  type 'a elm = 'a
+
+  let init length stream =
+    {stream; length}
+
+  let length t =
+    t.length
+
+  let next t =
+    match t.stream with
+    | lazy Stream.Nil -> not_reached ()
+    | lazy (Stream.Cons(elm, stream')) -> begin
+        let t' = {t with stream=stream'} in
+        elm, t'
+      end
+end
+
+module Array_of_stream = struct
+  include Array_of_stream_common
+  include Seq.Make_poly(Array_of_stream_common)
+end
+
+module Array_of_stream_rev = struct
+  include Array_of_stream_common
+  include Seq.Make_poly_rev(Array_of_stream_common)
+end
+
+let of_stream ?length stream =
+  let length = match length with
+    | None -> Stream.length stream
+    | Some length -> length
+  in
+  Array_of_stream.to_array (Array_of_stream.init length stream)
+
+let of_stream_rev ?length stream =
+  let length = match length with
+    | None -> Stream.length stream
+    | Some length -> length
+  in
+  Array_of_stream_rev.to_array (Array_of_stream_rev.init length stream)
+
 let is_empty t =
   (length t) = 0
 
@@ -1272,6 +1318,38 @@ let%expect_test "of_list,of_list_rev" =
     of_list[_rev] [0; 1] -> [|0; 1|] / [|1; 0|]
     of_list[_rev] [0; 1; 2] -> [|0; 1; 2|] / [|2; 1; 0|]
     of_list[_rev] [0; 1; 2; 3] -> [|0; 1; 2; 3|] / [|3; 2; 1; 0|]
+    |}]
+
+let%expect_test "of_stream,of_stream_rev" =
+  let open Format in
+  printf "@[<h>";
+  let test stream = begin
+    printf "of_stream[_rev] %a -> %a / %a\n"
+      (Stream.pp Usize.pp) stream
+      (pp Usize.pp) (of_stream stream)
+      (pp Usize.pp) (of_stream_rev stream);
+  end in
+  let f list = begin
+    match list with
+    | [] -> None
+    | hd :: tl -> Some (hd, tl)
+  end in
+  let streams = List.map ~f:(Stream.init_indef ~f) [
+    [];
+    [0];
+    [0; 1];
+    [0; 1; 2];
+    [0; 1; 2; 3];
+  ] in
+  List.iter streams ~f:test;
+  printf "@]";
+
+  [%expect{|
+    of_stream[_rev] Nil -> [||] / [||]
+    of_stream[_rev] (0 Nil) -> [|0|] / [|0|]
+    of_stream[_rev] (0 (1 Nil)) -> [|0; 1|] / [|1; 0|]
+    of_stream[_rev] (0 (1 (2 Nil))) -> [|0; 1; 2|] / [|2; 1; 0|]
+    of_stream[_rev] (0 (1 (2 (3 Nil)))) -> [|0; 1; 2; 3|] / [|3; 2; 1; 0|]
     |}]
 
 let%expect_test "get,length,is_empty" =
--- a/bootstrap/src/basis/file.c
+++ b/bootstrap/src/basis/file.c
@@ -1,0 +1,174 @@
+#include <errno.h>
+#include <fcntl.h>
+#include <stddef.h>
+#include <string.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <unistd.h>
+
+#include <caml/mlvalues.h>
+
+int flags_of_hemlock_file_flag[] = {
+  /* R_O   */  O_RDONLY,
+  /* W     */  O_WRONLY | O_CREAT,
+  /* W_A   */  O_WRONLY | O_APPEND | O_CREAT,
+  /* W_AO  */  O_WRONLY | O_APPEND,
+  /* W_C   */  O_WRONLY | O_CREAT | O_EXCL,
+  /* W_O   */  O_WRONLY,
+  /* RW    */  O_RDWR | O_CREAT,
+  /* RW_A  */  O_RDWR | O_APPEND| O_CREAT,
+  /* RW_AO */  O_RDWR | O_APPEND,
+  /* RW_C  */  O_RDWR | O_CREAT | O_EXCL,
+  /* RW_O  */  O_RDWR,
+};
+
+uint8_t * deflate_bytes(size_t n, size_t * inflated_bytes) {
+  uint8_t * deflated_bytes = (uint8_t *) inflated_bytes;
+  for (size_t i = 0; i < n; i++) {
+    deflated_bytes[i] = Int_val(inflated_bytes[i]);
+  }
+  return deflated_bytes;
+}
+
+size_t * inflate_bytes(size_t n, uint8_t * deflated_bytes) {
+  size_t * inflated_bytes = (size_t *) deflated_bytes;
+  for (size_t i = n; i > 0; i--) {
+    inflated_bytes[i - 1] = (size_t) Val_int(deflated_bytes[i - 1]);
+  }
+  return inflated_bytes;
+} 
+
+CAMLprim value
+hemlock_file_error_to_string_get_length(value a_error) {
+  int error = Int_val(a_error);
+
+  int length = strlen(strerror(error));
+
+  return Val_int(length);
+}
+
+CAMLprim value
+hemlock_file_error_to_string_inner(value a_n, value a_bytes, value a_error) {
+  size_t n = Long_val(a_n);
+  size_t * bytes = (size_t *) String_val(a_bytes);
+  int error = Int_val(a_error);
+
+  uint8_t * cbytes = (uint8_t *) bytes;
+  strncpy(cbytes, strerror(error), n);
+  inflate_bytes(n, cbytes);
+
+  return Val_unit;
+}
+
+CAMLprim value
+hemlock_file_finalize_result(int result) {
+  if (-1 == result) {
+    result = -errno;
+  }
+
+  return Val_int(result);
+}
+
+CAMLprim value
+hemlock_file_of_path_inner(value a_flag, value a_mode, value a_i, value a_j, value a_bytes) {
+  size_t flag = Long_val(a_flag);
+  size_t mode = Long_val(a_mode);
+  size_t i = Long_val(a_i);
+  size_t j = Long_val(a_j);
+  size_t * bytes = (size_t *) String_val(a_bytes);
+
+  int flags = flags_of_hemlock_file_flag[flag];
+  size_t n = j - i;
+  bytes = &bytes[i];
+  uint8_t * cbytes = deflate_bytes(n, bytes);
+  // TODO expose mode in API
+  int result = open(cbytes, flags, mode);
+  inflate_bytes(n, cbytes);
+
+  return hemlock_file_finalize_result(result);
+}
+
+CAMLprim value
+hemlock_file_stdin_inner(value a_unit) {
+  return Val_int(STDIN_FILENO);
+}
+
+CAMLprim value
+hemlock_file_stdout_inner(value a_unit) {
+  return Val_int(STDOUT_FILENO);
+}
+
+CAMLprim value
+hemlock_file_stderr_inner(value a_unit) {
+  return Val_int(STDERR_FILENO);
+}
+
+CAMLprim value
+hemlock_file_close_inner(value a_fd) {
+  int fd = Int_val(a_fd);
+
+  return hemlock_file_finalize_result(close(fd));
+}
+
+CAMLprim value
+hemlock_file_read_inner(value a_i, value a_j, value a_bytes, value a_fd) {
+  size_t i = Long_val(a_i);
+  size_t j = Long_val(a_j);
+  size_t * bytes = (size_t *) String_val(a_bytes);
+  int fd = Int_val(a_fd);
+
+  size_t n = j - i;
+  bytes = &bytes[i];
+  uint8_t * cbytes = (uint8_t *) bytes;
+  int result = read(fd, cbytes, n);
+  if (result > 0) {
+    inflate_bytes(result, cbytes);
+  }
+
+  return hemlock_file_finalize_result(result);
+}
+
+CAMLprim value
+hemlock_file_read_into_inner(value a_i, value a_j, value a_bytes, value a_fd) {
+  return hemlock_file_read_inner(a_i, a_j, a_bytes, a_fd);
+}
+
+CAMLprim value
+hemlock_file_write_inner(value a_i, value a_j, value a_bytes, value a_fd) {
+  size_t i = Long_val(a_i);
+  size_t j = Long_val(a_j);
+  size_t * bytes = (size_t *) String_val(a_bytes);
+  int fd = Int_val(a_fd);
+
+  size_t n = j - i;
+  bytes = &bytes[i];
+  uint8_t * cbytes = deflate_bytes(n, bytes);
+  int result = write(fd, cbytes, n);
+  inflate_bytes(n, cbytes);
+
+  return hemlock_file_finalize_result(result);
+}
+
+CAMLprim value
+hemlock_file_seek_inner(value a_i, value a_fd) {
+  size_t i = Long_val(a_i);
+  int fd = Int_val(a_fd);
+
+  return hemlock_file_finalize_result(lseek(fd, i, SEEK_CUR));
+}
+
+CAMLprim value
+hemlock_file_seek_hd_inner(value a_i, value a_fd) {
+  size_t i = Long_val(a_i);
+  int fd = Int_val(a_fd);
+
+  return hemlock_file_finalize_result(lseek(fd, i, SEEK_SET));
+}
+
+CAMLprim value
+hemlock_file_seek_tl_inner(value a_i, value a_fd) {
+  size_t i = Long_val(a_i);
+  int fd = Int_val(a_fd);
+
+  return hemlock_file_finalize_result(lseek(fd, i, SEEK_END));
+}
--- a/bootstrap/src/basis/file.ml
+++ b/bootstrap/src/basis/file.ml
@@ -1,0 +1,246 @@
+open Rudiments
+
+module Error = struct
+  type t = usize
+
+  (* external to_string_get_length: t -> usize *)
+  external to_string_get_length: t -> usize = 
+    "hemlock_file_error_to_string_get_length"
+
+  (* external to_string_inner: usize -> Bytes.t -> t $-> unit *)
+  external to_string_inner: usize -> Bytes.t -> t -> unit =
+    "hemlock_file_error_to_string_inner"
+
+  let of_value value =
+    Usize.of_isize Isize.(-value)
+
+  let to_string t =
+    let n = to_string_get_length t in
+    let bytes = Array.init n ~f:(fun _ -> Byte.of_usize 0) in
+    let () = to_string_inner n bytes t in
+    Bytes.to_string_hlt bytes
+end
+
+module Flag = struct
+  (* Modifications to Flag.t must be reflected in file.c. *)
+  type t =
+    | R_O
+    | W
+    | W_A
+    | W_AO
+    | W_C
+    | W_O
+    | RW
+    | RW_A
+    | RW_AO
+    | RW_C
+    | RW_O
+end
+
+module Buffer = struct
+  (* type /t t *)
+  type t = {
+    (* hd: byte /t Array.Cursor.t *)
+    hd: Byte.t Array.Cursor.t;
+    (* hd: byte /t Array.Cursor.t *)
+    tl: Byte.t Array.Cursor.t;
+  }
+
+  let of_bytes bytes =
+    let hd = Array.Cursor.hd bytes in
+    let tl = Array.Cursor.tl bytes in
+    {hd; tl}
+
+  let of_string string =
+    of_bytes (Bytes.of_string string)
+
+  (* i: /t t -> usize *)
+  let i t =
+    Array.Cursor.index t.hd
+
+  (* j: /t t -> usize *)
+  let j t =
+    Array.Cursor.index t.tl
+
+  (* container: /t t -> Byte.t /t Array.t *)
+  let container t =
+    Array.Cursor.container t.hd
+
+  (* length: /t t -> usize *)
+  let length t =
+    (Array.Cursor.index t.tl) - (Array.Cursor.index t.hd)
+end
+
+type t = usize
+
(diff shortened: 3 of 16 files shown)


cevans87:
The `File` module as is uses a bit less exception handling than would be needed for an actual robust implementation build on top of the Unix module. I actually can't get most of the kinds of errors I actually expect in the Hemlock implementation. I'm wondering if I should write small wrappers around the `Unix` module calls so that they give back the c-like error codes I'd expect in Hemlock.

I think there are a few nitpicky problems with function implementations as well. Go ahead and point them out if you see them, but right now I'm more worried about making this look translatable into Hemlock.

bootstrap/src/basis/file.ml
```diff
@@ -0,0 +1,243 @@
...
+let of_path ?(mode=Mode.RW) path =
+  let open Unix in
+  let unix_flags = Mode.to_unix_flags mode in
+  try
+    Ok ({fd=Some (openfile path unix_flags 0o640)})
+  with
+  | Unix_error (errno, fn_name, fn_param) -> begin
+    Error (String.concat ~sep:" -> " [Unix.error_message errno; fn_name; fn_param])
+  end
+  | _ -> not_reached ()
```
cevans87, line 52: So, this is a good example of where I'd like to improve things. The contents of this function look nothing like what Hemlock will look like. I'd like to hide the `Unix.openfile` call and exception handler behind a wrapper that looks more Hemlock-like, but not really sure what the wrapper function signature should look like. I also wonder if it'd be easier to wrap native c function calls than to wrap `Unix` calls.

bootstrap/src/basis/file.ml
```diff
@@ -0,0 +1,243 @@
...
+
+let%expect_test "Stream.of_path_hlt" =
+  let open Format in
+  printf "@[<h>";
+  let _ = Stream.of_path_hlt "/home/cevans/test.txt"
+    |> Stream.write_hlt (of_path_hlt ~mode:Mode.W_A "/home/cevans/test2.txt")
+   in
+  printf "@]";
+
+  [%expect{| |}]
```
cevans87, line 243: I'll get rid of this.

bootstrap/src/basis/file.ml
jasone: Writing native wrappers around the system calls sounds like a pretty good idea, actually.  The result would be quite close to how the Hemlock code will look.

bootstrap/src/basis/file.ml
jasone: Note that it's possible to use the stand-alone test infrastructure similar to what I did for the entropy tests (in `bootstrap/test/basis/`) for the cases where you need a test file to operate on.

bootstrap/src/basis/file.mli
```diff
@@ -0,0 +1,122 @@
...
+  | RW_O  (** truncate existing file and open for read and write, failing if
+              file does not exist *)
+end
+
+type t
+(** mutable Unix file descriptor *)
+
+type e
+
+val of_path: ?mode:Mode.t -> Path.t -> (t, e) Result.t
```
jasone, line 30: It's probably worthwhile documenting what the default for `mode` is.

bootstrap/src/basis/file.mli
```diff
@@ -0,0 +1,122 @@
...
+val read_hlt: ?n:usize -> t -> Bytes.t
+(** [read_hlt n t] reads up to [n] bytes from [t] and returns the read bytes or
+    halts if bytes could not be read. *)
+
+val write: Bytes.t -> t -> e option
+(** [write bytes t] writes [bytes] to [t] and returns an error if bytes could
+    not be written. *)
+
+val write_hlt: Bytes.t -> t -> unit
+(** [write_htl bytes t] writes [bytes] to [t] and halts if bytes could not be
```
jasone, line 51:
```suggestion
(** [write_hlt bytes t] writes [bytes] to [t] and halts if bytes could not be
```

bootstrap/src/basis/file.mli
```diff
@@ -0,0 +1,122 @@
...
+
+val of_path: ?mode:Mode.t -> Path.t -> (t, e) Result.t
+(** [of_path mode path] opens file in [mode] at [path] and returns the resulting
+    file or an error if file could not be opened. *)
+
+val of_path_hlt: ?mode:Mode.t -> Path.t -> t
+(** [of_path mode path] opens file in [mode] at [path] and returns the resulting
+    file or halts if file could not be opened. *)
+
+val read: ?n:usize -> t -> (Bytes.t, e) Result.t
```
jasone, line 38: What do you think about taking a mutable `bytes` parameter here?  Otherwise we have to allocate an array every time we read.  Or maybe we want both APIs, so that we can minimize copying for each case?

bootstrap/src/basis/file.ml
```diff
@@ -0,0 +1,243 @@
...
+    end in
+    fn file t
+
+  let write_hlt file t =
+    match write file t with
+    | Some e -> halt e
+    | None -> ()
+
+end
+
```
jasone, line 207:
```suggestion

(******************************************************************************)
(* Begin tests. *)
```

bootstrap/src/basis/file.mli
```diff
@@ -0,0 +1,122 @@
...
+val seek: isize -> t -> (usize, e) Result.t
+(** [seek i t] mutates file descriptor in [t] to point to the [i]th byte
+    relative to the current byte position of the file. Returns the new byte
+    index relatve to the beginning of the file or an error if it could not be
+    changed. *)
+
+val seek_hlt: isize -> t -> usize
+(** [seek_hlt i t] mutates file descriptor in [t] to point to the [i]th byte
+    relative to the current byte position of the file. Returns the new byte
+    index relatve to the beginning of the file or halts if it could not be
```
jasone, line 70:
```suggestion
    index relative to the beginning of the file or halts if it could not be
```

jasone: This typo is in several places.

bootstrap/src/basis/file.mli
```diff
@@ -0,0 +1,122 @@
...
+
+val seek_left: usize -> t -> (usize, e) Result.t
+(** [seek_left u t] mutates file descriptor in [t] to point to the [u]th byte
+    after to the beginning of the file. Returns the new byte index relatve to
+    the beginning of the file or an error if it could not be changed. *)
+
+val seek_left_hlt: usize -> t -> usize
+(** [seek_left_hlt u t] mutates file descriptor in [t] to point to the [u]th
+    byte after to the beginning of the file. Returns the new byte index relatve
+    to the beginning of the file or an error if it could not be changed. *)
```
jasone, line 81:
```suggestion
    to the beginning of the file or halts if it could not be changed. *)
```

jasone: Same for `seek_right_hlt`.

bootstrap/src/basis/file.ml
```diff
@@ -0,0 +1,243 @@
...
+let seek_right u t =
+  let i = isize_of_usize u in
+  _seek Unix.SEEK_END i t
+
+let seek_right_hlt u t =
+  let i = isize_of_usize u in
+  _seek_hlt Unix.SEEK_END i t
+
+module Stream = struct
+
```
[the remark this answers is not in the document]
jasone: Same elsewhere.

bootstrap/src/basis/file.mli
```diff
@@ -0,0 +1,122 @@
...
+
+module Stream : sig
+
+  type outer = t
+  type t = byte Stream.t
+
+  val of_file: outer -> t
+  (** [of_file file] takes an open [file] with read permissions and returns a
+      byte stream of the contents. Reading from the returned byte stream mutates
+      the read position of [file]. may cause the returned byte stream to  *)
```
jasone, line 101: ?

bootstrap/src/basis/file.ml
```diff
@@ -0,0 +1,243 @@
...
+    Ok ({fd=Some (openfile path unix_flags 0o640)})
+  with
+  | Unix_error (errno, fn_name, fn_param) -> begin
+    Error (String.concat ~sep:" -> " [Unix.error_message errno; fn_name; fn_param])
+  end
+  | _ -> not_reached ()
+
+let of_path_hlt ?mode path =
+  let t_res = of_path ?mode path in
+  match t_res with
```
jasone, line 56:
```suggestion
  match of_path ?mode path with
```

bootstrap/src/basis/file.ml
```diff
@@ -0,0 +1,243 @@
...
+  | Ok bytes -> bytes
+
+let write bytes t =
+  match t.fd with
+  | None -> Some "File is closed"
+  | Some fd ->
+    let rec fn i n chars fd = begin
+      let n' = Unix.write fd chars i n in
+      match n' = -1 with
+      | true -> Some "Error reading file"
```
jasone, line 81:
```suggestion
      | true -> Some "Error writing file"
```

bootstrap/src/basis/file.mli
```diff
@@ -0,0 +1,122 @@
...
+
+val read: ?n:usize -> t -> (Bytes.t, e) Result.t
+(** [read n t] reads up to [n] bytes from [t] and returns the read bytes or an
+    error if bytes could not be read. *)
+
+val read_hlt: ?n:usize -> t -> Bytes.t
+(** [read_hlt n t] reads up to [n] bytes from [t] and returns the read bytes or
+    halts if bytes could not be read. *)
+
+val write: Bytes.t -> t -> e option
```
jasone, line 46: Maybe we need to support partial writes.  Ordinarily that just causes a bunch of trouble for files (yay interruptible syscalls), but what about for broken pipes, where we did write some bytes before the pipe was closed?  I think this comes up for sockets too.

```ocaml
val write: Bytes.t -> t -> (usize, e) Result.t
```

bootstrap/src/basis/file.ml
jasone: Same basic idea in numerous other places.  As a general rule, if you have to make up a temporary name that doesn't aid understanding, inline the expression.

bootstrap/src/basis/file.ml
```diff
@@ -0,0 +1,243 @@
...
+        | Some error -> Some error
+      end
+    end in
+    fn file t
+
+  let write_hlt file t =
+    match write file t with
+    | Some e -> halt e
+    | None -> ()
+
```
jasone, line 205:
```suggestion
```

bootstrap/src/basis/file.ml
```diff
@@ -0,0 +1,243 @@
...
+    | Error e -> halt e
+    | Ok(t) -> t
+
+  let write file t =
+    let rec fn file t = begin
+      match t with
+      | lazy Stream.Nil -> None
+      | lazy (Stream.Cons(_, _)) -> begin
+        let t', t'' = Stream.split buflen t in
+        let bytes = Array.of_stream t' in
```
jasone, line 192: So nice.

cevans87:
Cool, I agree about dropping the `Unix` module and try to do this with native c wrappers. `File.ml` will look quite different after I revise, but I think it'll work much better as a bootstrap implementation.

I'll fix the doc comments and add real file I/O tests in `bootstrap/test/basis/`.

bootstrap/src/basis/array.ml
```diff
@@ -1274,6 +1320,38 @@ let%expect_test "of_list,of_list_rev" =
...
+      (Stream.pp Usize.pp) stream
+      (pp Usize.pp) (of_stream stream)
+      (pp Usize.pp) (of_stream_rev stream);
+  end in
+  let f list = begin
+    match list with
+    | [] -> None
+    | hd :: tl -> Some (hd, tl)
+  end in
+  let streams = List.map ~f:(Stream.init_indef ~f) [
```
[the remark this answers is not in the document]
cevans87: I was tempted to add `Stream.of_list`, but I'll leave that alone for now and take care of it when I try to put `Stream` everywhere that it belongs.

bootstrap/src/basis/file.mli
cevans87: You're right. Maybe partial writes should return an error that contain the number of lines written, and `None` still indicates success? Regardless, the unix standard write operation returns the number of bytes written, so we should as well.

bootstrap/src/basis/file.mli
cevans87: It depends on what I expect to know from the OS. If we can't know how many bytes the OS has available for us, then we should take a mutable `bytes`. If we were building this on top of poll or epoll, I think that'd be the case. I'll check what kind of info we'll expect from io_uring. If we know how many bytes are available before reading, we could do a better job of constructing small buffers for small reads right before we actually do the read. I need to do a little research on this before I can decide.

bootstrap/src/basis/file.mli
```diff
@@ -0,0 +1,122 @@
...
+
+val seek_right: usize -> t -> (usize, e) Result.t
+(** [seek_right u t] mutates file descriptor in [t] to point to the [u]th byte
+    before to the end of the file. Returns the new byte index relatve to the
+    beginning of the file or an error if it could not be changed. *)
+
+val seek_right_hlt: usize -> t -> usize
+(** [seek_right_hlt u t] mutates file descriptor in [t] to point to the [u]th
+    byte before to the end of the file. Returns the new byte index relatve
+    to the beginning of the file or an error if it could not be changed. *)
```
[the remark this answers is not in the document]
cevans87: I based this on the behavior of `fold_right` that I saw somewhere else in `Basis`. I actually had the exact same misunderstanding with `fold_right`. However, `hd_seek`, `tl_seek`, would be clearer.

jasone:
Oh, one thing we're missing here is a way to access files which have names that aren't UTF-8.  We talked about having a bottom-level open function which takes a `Bytes.t` rather than`string` for the path.

cevans87:
> Oh, one thing we're missing here is a way to access files which have names that aren't UTF-8. We talked about having a bottom-level open function which takes a `Bytes.t` rather than`string` for the path.

Ah, right. I'll fix that. I was hasty here and just plugged in the type that OCaml didn't complain about. Iirc, most filesystems just care about the byte content of the filename, but shells often interpret the bytes as utf8 depending on the locale environment variables?

jasone:
> Iirc, most filesystems just care about the byte content of the filename, but shells often interpret the bytes as utf8 depending on the locale environment variables?

Yep, that's the way I understand it.  Dealing with non-UTF-8 in the shells is difficult these days, which means non-UTF-8 file names almost never come up in practice.  But we can't completely ignore them, unfortunately.

bootstrap/src/basis/file.mli
```diff
@@ -0,0 +1,156 @@
...
+val seek_tl_hlt: ?i:isize -> t -> t
+(** [seek_right_hlt n t] mutates file descriptor in [t] to point to the [u]th
+    byte before to the end of the file. Returns the new byte index relatve
+    to the beginning of the file or an error if it could not be changed. *)
+
+module Stream : sig
+  type file = t
+  type t = Buffer.t Stream.t
+
+  (* val read: ?file !-> file * (Buffer.t, Error.t) result Stream.t *)
```
jasone, line 138:
```suggestion
  (* val of_file: file -> (Buffer.t, Error.t) result Stream.t *)
```
I'm uneasy about requiring isolated file references here (?file), because it requires some care to keep a reference isolated until it is consumed. Additionally, there's no benefit here, since we return a reference to the file.

bootstrap/src/basis/file.mli
```diff
@@ -0,0 +1,156 @@
...
+  val of_file: file -> file * (Buffer.t, Error.t) result Stream.t
+
+  (* val read_hlt: ?file !-> file * t *)
+  val of_file_hlt: file -> file * t
+
+  (* val write: ?file -> t !-> Error.t option * ?file  *)
```
jasone, line 144:
```suggestion
  (* val write: file -> t !-> Error.t option *)
```

bootstrap/src/basis/file.mli
```diff
@@ -0,0 +1,156 @@
...
+val of_path: ?flag:Flag.t -> ?mode:usize -> Bytes.t -> (t, Error.t) result
+(** [of_path mode path] opens file in [mode] at [path] and returns the resulting
+    file or an error if file could not be opened. *)
+
+(* val of_path_hlt: Flag.t -> usize -> Bytes.t !-> ?t *)
+val of_path_hlt: ?flag:Flag.t -> ?mode:usize -> Bytes.t -> t
+(** [of_path mode path] opens file in [mode] at [path] and returns the resulting
+    file or halts if file could not be opened. *)
+
+(* val of_stdin: unit !-> Bytes.t !-> ?t *)
```
jasone, line 49:
```suggestion
(* val of_stdin: unit !-> Bytes.t !-> t *)
```
The `file` isn't necessarily isolated; it could be a copy of a value in the environment.  Same for `stdout` and `stderr`.  In fact, these should probably be much simpler (values rather than functions).
```suggestion
(* val stdin: t *)
```

bootstrap/src/basis/file.c
```diff
@@ -0,0 +1,167 @@
...
+  // TODO expose mode in API
+  int result = open(cbytes, flags, mode);
+  inflate_bytes(n, cbytes);
+
+  return hemlock_file_finalize_result(result);
+}
+
+CAMLprim value
+hemlock_file_of_stdin_inner(value a_unit) {
+  return hemlock_file_finalize_result(dup(STDIN_FILENO));
```
jasone, line 91: We shouldn't dup the terminal file descriptors.

bootstrap/src/basis/file.mli
cevans87: My intent here was to consume a unique reference to a file. The stream would own it. The `file` that we return here would only be useful for the `File.close` function, basically the interruptable type we mentioned before. Every other function in the API that takes a file input except `File.close` requires a the file to be a unique reference.

jasone: Unique references are rather precious, and they require meticulous care.  I've been calling them "isolated" rather than "unique", because the only thing that keeps them unique is isolation.  We should definitely chat about this in more detail and figure out whether they're appropriate for this use case.

bootstrap/src/basis/file.mli
```diff
@@ -0,0 +1,163 @@
...
+
+(* val stderr: t *)
+val stderr: t
+
+(* val of_path: ?flag:Flag.t -> ?mode:usize -> ^Buffer.t $-> (t, Error.t) result *)
+val of_path: ?flag:Flag.t -> ?mode:usize -> Buffer.t -> (t, Error.t) result
+(** [of_path mode path] opens file in [mode] at [path] and returns the resulting
+    file or an error if file could not be opened. *)
+
+(* val of_path_hlt: ?flag:Flag.t -> ?mode:usize -> ^Buffer.t !-> t *)
```
jasone, line 63:
```suggestion
(* val of_path_hlt: ?flag:Flag.t -> ?mode:usize -> ^Buffer.t !$-> t *)
```

bootstrap/src/basis/file.mli
```diff
@@ -0,0 +1,163 @@
...
+val of_path_hlt: ?flag:Flag.t -> ?mode:usize -> Buffer.t -> t
+(** [of_path mode path] opens file in [mode] at [path] and returns the resulting
+    file or halts if file could not be opened. *)
+
+(* val read_into: !^Buffer.t -> t $-> Error.t option *)
+val read_into: Buffer.t -> t -> (Buffer.t, Error.t) result
+
+(* val read_into_hlt: !^Buffer.t -> t !-> Buffer.t *)
```
jasone, line 71:
```suggestion
(* val read_into_hlt: !^Buffer.t -> t !$-> Buffer.t *)
```

bootstrap/src/basis/file.mli
```diff
@@ -0,0 +1,163 @@
...
+val read_into_hlt: Buffer.t -> t -> Buffer.t
+
+(* val read: t $-> (@Buffer.t, Error.t) result *)
+val read: t -> (Buffer.t, Error.t) result
+(** [read n t] reads up to [n] bytes from [t] and returns the read bytes or an
+    error if bytes could not be read. *)
+
+(* val read_hlt: t !-> @Buffer.t *)
```
jasone, line 79:
```suggestion
(* val read_hlt: t !$-> @Buffer.t *)
```

bootstrap/src/basis/file.mli
```diff
@@ -0,0 +1,163 @@
...
+val read_hlt: t -> Buffer.t
+(** [read_hlt n t] reads up to [n] bytes from [t] and returns the read bytes or
+    halts if bytes could not be read. *)
+
+(* val write: ^Buffer.t -> t $-> (usize, Error.t) result *)
+val write: Buffer.t -> t -> Error.t option
+(** [write bytes t] writes [bytes] to [t] and returns an error if bytes could
+    not be written. *)
+
+(* val write_hlt: ^Buffer.t -> t !-> usize *)
```
jasone, line 89:
```suggestion
(* val write_hlt: ^Buffer.t -> t !$-> usize *)
```

bootstrap/src/basis/file.mli
```diff
@@ -0,0 +1,163 @@
...
+val write_hlt: Buffer.t -> t -> unit
+(** [write_htl bytes t] writes [bytes] to [t] and halts if bytes could not be
+    written. *)
+
+(* val close: t $-> Error.t option *)
+val close: t -> Error.t option
+(** [close t] closes file [t] and returns an error if it could not be closed.
+    *)
+
+(* val close_hlt: t !-> unit *)
```
jasone, line 99:
```suggestion
(* val close_hlt: t !$-> unit *)
```

bootstrap/src/basis/file.mli
```diff
@@ -0,0 +1,163 @@
...
+(** [close_hlt t] closes file [t] and halts if it could not be closed. *)
+
+(* val seek: ?i:isize -> t $-> (usize, Error.t) result *)
+val seek: ?i:isize -> t -> (usize, Error.t) result
+(** [seek n t] mutates file descriptor in [t] to point to the [i]th byte
+    relative to the current byte position of the file. Returns the new byte
+    index relatve to the beginning of the file or an error if it could not be
+    changed. *)
+
+(* val seek_hlt: isize -> t !-> usize *)
```
jasone, line 110:
```suggestion
(* val seek_hlt: isize -> t !$-> usize *)
```

bootstrap/src/basis/file.mli
```diff
@@ -0,0 +1,163 @@
...
+    index relatve to the beginning of the file or halts if it could not be
+    changed. *)
+
+(* val seek_hd: isize -> t $-> (usize, Error.t) result *)
+val seek_hd: ?i:isize -> t -> (usize, Error.t) result
+(** [seek_left i t] mutates file descriptor in [t] to point to the [u]th byte
+    after to the beginning of the file. Returns the new byte index relatve to
+    the beginning of the file or an error if it could not be changed. *)
+
+(* val seek_hd_hlt: isize -> t !-> usize *)
```
jasone, line 123:
```suggestion
(* val seek_hd_hlt: isize -> t !$-> usize *)
```

bootstrap/src/basis/file.mli
```diff
@@ -0,0 +1,163 @@
...
+    byte after to the beginning of the file. Returns the new byte index relatve
+    to the beginning of the file or an error if it could not be changed. *)
+
+(* val seek_tl: ?i:isize -> t $-> (usize, Error.t) result *)
+val seek_tl: ?i:isize -> t -> (usize, Error.t) result
+(** [seek_right i t] mutates file descriptor in [t] to point to the [u]th byte
+    before to the end of the file. Returns the new byte index relatve to the
+    beginning of the file or an error if it could not be changed. *)
+
+(* val seek_tl_hlt: ?i:isize -> t !-> usize *)
```
jasone, line 135:
```suggestion
(* val seek_tl_hlt: ?i:isize -> t !$-> usize *)
```

bootstrap/src/basis/file.mli
```diff
@@ -0,0 +1,163 @@
...
+  (* val of_file: file $-> t *)
+  val of_file: file -> t
+
+  (* val write: file -> t $-> Error.t option  *)
+  val write: file -> t -> Error.t option
+  (** [write file byte_stream] takes an open [file] with write permissions and
+    writes contents of [byte_stream] to it. Returns an error if not all bytes
+    could be written. *)
+
+  (* val write_hlt: file -> t !-> unit  *)
```
jasone, line 157:
```suggestion
  (* val write_hlt: file -> t !$-> unit  *)
```

bootstrap/src/basis/stream.mli
```diff
@@ -34,63 +38,76 @@ val init_indef: f:('state -> ('a * 'state) option) -> 'state -> 'a t
...
+(* val pop: 'a >e t -> 'a * 'a >e t *)
 val pop: 'a t -> 'a * 'a t
 (** Pop head element off stream and return the decomposed element and remainder
     stream or halt if the stream is empty. *)
 
+(* val concat: 'a >e t -> 'a >e t -> 'a >e t *)
 val concat: 'a t -> 'a t -> 'a t
 (** Concatenate two streams. *)
 
+(* val split: usize -> 'a >e t -> 'a >e t * 'a >e t *)
```
jasone, line 76: `split` and `rev_split` potentially have to force in order to find the split point, right? That means `>e->` for the rightmost arrow. Actually, I think this comment applies to all the following functions.

bootstrap/src/basis/stream.mli
```diff
@@ -34,63 +38,76 @@ val init_indef: f:('state -> ('a * 'state) option) -> 'state -> 'a t
...
 val length: 'a t -> usize
 (** Return stream length.  This forces and counts every node and is Θ(n). *)
 
+(* val is_empty: 'a >e t >e-> bool  *)
 val is_empty: 'a t -> bool
 (** Return [true] if stream length is 0; [false] otherwise. *)
 
 (** {1 Element access} *)
 
+(* val hd: 'a >e t >e-> 'a *)
```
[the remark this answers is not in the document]
cevans87: `hd` definitely can't. It has to force to get the a value out of the stream. `tl` could be effectless, but the implementation actually forces once to get the tail.

bootstrap/src/basis/stream.mli
cevans87: I had to look at the implementations to see what actually happens. No work is actually done until one of the streams is forced.

jasone:
Merged via #84.

PR #75, opened 2020-07-01
Begin adding design documentation.

jasone:
There are numerous `XXX` annotations where further design documentation is warranted.  Still this is a big start, the culmination of nearly three months of effort.

### DIFF
--- a/doc/design/effects_mutability.md
+++ b/doc/design/effects_mutability.md
@@ -1,0 +1,334 @@
+# Effects and mutability
+
+If Hemlock were a pure functional language, it would be effectless; its output
+would depend only on its initial input. Hemlock does have a pure, effectless
+core, but non-trivial Hemlock programs rely on effects, which can be roughly
+categorized as:
+
+- Mutation of mutable data structures. For example, mutable arrays are sometimes
+  the most efficient data structure for highly interdependent data and
+  algorithms.
+- Messages between actors. A Hemlock program need not run more than a single
+  actor, but managing actor lifetimes, sending/receiving messages to/from other
+  actors, etc. is all effectful computation.
+- Input/output mediated by the host operating system, whether for durable
+  storage, communication with other processes and/or computers, etc., are all
+  program-external effects.
+
+Internal effects can be thought of entirely in terms of reading/writing mutable
+program values. Reading a mutable value may not seem like it should be
+considered an effect, but the fact that the mutable value does not have a stable
+state means that repeated reads of the same may return differing results, thus
+affecting subsequent computation. Effectful reads must happen at "the right
+time" to preserve program correctness, whereas reads of immutable values may
+occur at any time prior to use.
+
+External effects are similar to internal effects in that repeating an effectful
+operation may produce differing results. However, Hemlock cannot model external
+state transitions with as much detail as for internal effects, so no practical
+distinction can be made between external read versus write effects.
+
+## Syntax quick reference
+
+For more semantics detail than is provided in the following syntax quick
+reference, see below.
+
+- Effects
+  - `>e`: Parametric effect.
+  - `>_`: Effects are contextually irrelevant, and therefore arbitrarily
+    permitted.
+  - `.`: Internal environment read effect.
+  - `!`: Internal environment read-write effect if used as e.g. `!->`; parameter
+    mutation effect if used as e.g. `!&t`.
+  - `$`: External environment read-write effect.
+  - `<.`, `<!`, `<$`, `<.$`, `<!$`: Effects constraint: transitively prohibited
+    effect(s).
+- Mutability
+  - `/m`: Parametric mutability.
+  - `/_`: Mutability is contextually irrelevant, and therefore both immutable
+    and mutable values are supported.
+  - `&`: Non-parametric transitive mutability, whether direct and/or indirect.
+    `!&t` indicates an effect on the value, which must therefore be mutable.
+  - `/&`: Mutability constraint: transitively required mutability.
+  - `\&`: Mutability constraint: transitively prohibited mutability.
+- Parametric types:
+  - `'a`: Parametric type, with implicit effects and mutability parameters,
+    equivalent to `>a &a 'a`.
+  - `'a &t`: Parametric type with transitive mutability.
+  - `val f: 'a -> 'a /t t`: The function produces an isolated value with
+    parametric mutability. The value's mutability is either immutable or mutable
+    as determined by type unification in the caller's context.
+  - `('a, /m, >e) ['c, 'd] &t`: Mutable type with transitive effects and
+    mutability determined by `('a, /m, >e)`, and additional parameters which
+    have no impact on the type's transitive effects and mutability determined by
+    `['c, 'd]`.
+  - `'a <. \& t`: Parametric type constrained to exclude internal effects and
+    mutability.
+
+# Static inference
+
+Hemlock statically determines all effects and mutability transitively associated
+with a type, whether direct or indirect. Internal effects and mutability are
+prohibited in global and message values, though external effects (i.e. outside
+the Hemlock execution environment) are allowed. Internal effects are lexically
+bound to a particular actor's local heap, and therefore cannot be supported in
+the global heap. These restrictions are evaluated at compile time, which means
+that the programming model is unaffected by type erasure. The absence of dynamic
+validation overhead enables fast message passing, as well as fully automated
+migration of compatible long-lived values to the global heap.
+
+# Parametrization
+
+Non-parametric types have fixed effects and mutability, whereas parametric types
+also have transitive effects and/or mutability that allow differing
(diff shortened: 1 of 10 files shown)


jasone:
The latest update contains only formatting changes.  (Yay for proper automated rewrapping!)

doc/design/effects_mutability.md
```diff
@@ -0,0 +1,266 @@
...
+
+Non-parametric types have fixed effects and mutability, whereas parametric types
+also have transitive effects and/or mutability that allow differing
+parametrizations to be any combination of effectless/effectful and/or
+immutable/mutable.  For example, `'a array` is directly immutable, but it may be
+transitively mutable depending on the particulars of the type supplied to the
+`'a` parameter.  Were we to attempt passing `'a array` in a message, the
+compiler would be unable to prove that the message is compatible with the
+aforementioned constraints.  Therefore we must specify transitive constraints in
+the type, e.g. `'a <.  \& array`.  Similarly, finalization is only supported for
```
cevans87, line 21: In pretty much all the cases where we get a sneak peak at type specifier syntax, I skipped to the end of the doc to reference the meaning and wondered whether I'd be really confused had I not yet known the summary exists. I wonder if it'd be better to have the summary at the beginning. Maybe call it a "quick reference" if that's more appropriate since it no longer comes after the deeper explanations.

doc/design/effects_mutability.md
```diff
@@ -0,0 +1,266 @@
...
+be explicitly written as `>a /a 'a`.  For example, the set type could be written
+as `(>a /a\& 'a, >cmp /cmp 'cmp) set`.  The `/a` parameter is constrained to
+prohibit mutability, lest key mutation corrupt the set.
+
+Regarding effects versus mutability consider that `'a &array` is a mutable type,
+but actual mutation is an effect.
+
+```hemlock
+val length: 'a /_ array -> usize
+val set: 'a -> usize -> 'a !&array -> unit
```
cevans87, line 54: I don't think I ever fully grasped what counts as an internal read-write effect. What I wonder is if mutating a functions input counts as an internal write effect.
```ocaml
let foo arr = begin
  let bar arr () = begin
    Array.set_inplace 0 42 arr
  end in
  bar arr
```
Would the type signature here be
```Ocaml
val foo: uns !&array -> (unit !-> unit)
```
The reason I ask is that I think it might matter that we specify where the mutable input actually gets mutated. Otherwise, that type signature I just showed might be
```Ocaml
val foo: uns !&array -> (unit -> unit)
```
Which clearly has some problems. I think the example you show on this line has the same problem, and I'm wondering if it's basically a typo or if we missed something in the type system.

Also, I'm guessing we'll have global values. Do I have some of these examples correct?

Global read implementation.
```Ocaml
let arr = Array.init 42 f:(i -> i)
let foo () = begin
  arr[0]
end
```
Global read signature.
```Ocaml
val foo: unit .-> uns
```

Global write implementation.
```Ocaml
let arr = Array.init 42 f:(i -> i)
let foo () = begin
  let () = Array.set_inplace 0 1 arr in
  arr[0]
end
```
Global write signature.
```Ocaml
val foo: unit !-> uns
```

jasone: I added a section regarding partial application and (not) losing track of effects.  In short, the first `foo` example you gave does need a `!->` arrow in the resulting closure, and there's a principled way to always get this right.

I also rearranged things and added a gentler intro to effects and mutability.  Let me know if it still needs help to be straightforward reading.

doc/design/effects_mutability.md
jasone: Good idea; done.

doc/design/memory.md
```diff
@@ -0,0 +1,1595 @@
...
+
+  cccccccc cccccccc cccccccc cccccccc cccccccc cccccccc cccccccc ccccc0z1
+
+Forwarding pointer:
+
+  ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff fffff000
+
+Zero refcount, deallocated, in space's unreachable value stack:
+
+  ssssssss ssssssss ssssssss ssssssss ssssssss ssssssss ssssssss sssss0z0
```
cevans87, line 476: Just to make sure I understand this. Whenever a value is decrefed, to zero, we take over the `s` value portion of this header and make it point to the old top of the unreachable value stack. At the same time, we make the space metadata header point to this current header as the top of the unreachable value stack? At some point, the GC will come along and evaluate the actual value contents associated with this header to figure out if it points to any other values it needs to decref. Once the GC is done with that, it zeros out this entire header.

doc/design/memory.md
```diff
@@ -0,0 +1,1595 @@
...
+lg(quantum) address bits, and each space typically only requires one entry in
+the radix tree.  (The extremely unusual exception is for a multi-quantum space
+that contains a huge value, and interior references to that value allow
+references to span multiple quanta.)
+
+### Generation coalescing
+
+The nursery generation grows until its space cannot serve an allocation request,
+at which point it becomes an aging generation.  Aside from allocation via lazy
+suspension execution, a generation cannot grow after it transitions to aging.
```
cevans87, line 502: What do we do if there's enough data in the lazy space to require more than one space after copy/collect?

doc/design/memory.md
```diff
@@ -0,0 +1,1595 @@
...
+There is no obvious fundamental principle on which to base pace of advancement,
+except perhaps nursery creation rate and/or impending memory exhaustion.  On the
+other hand, decrefing and copying can be driven at a rate proportional to
+allocation rate that is sufficient to guarantee this work does not fall
+hopelessly behind allocation.
+
+Copying is triggered at a fixed utilization threshold, so the mean number of
+times a value can be copied is fixed.  For example, if minimum utilization is
+0.5, the total number of bytes copied per byte allocated is approximately the
+integral of 0.5<sup>x</sup> in the range 0..∞, which is ~1.44.  Therefore the
```
cevans87, line 541: Neat, I'd actually been wondering how much we could possibly copy. It's cool that the answer was so simple.

doc/design/memory.md
```diff
@@ -0,0 +1,1595 @@
+# Memory management
+
+Hemlock depends on automatic memory management, also known as garbage collection
+(GC).  The memory management subsystem implements precise tracking of memory
+references, and its GC algorithms are capable of collecting all unreachable
+data, given sufficient time.
+
+Hemlock's memory management subsystem has to support parallel execution of as
+many runtime executors as there are CPUs, and on the order of millions of actors
+multiplexed onto those executors.  The GC algorithms employed by Hemlock (and
```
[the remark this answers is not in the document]
jasone: Oh man. I went and researched inter-sentence spacing this morning, and learned a lot. Both sides of the issue are commonly couched in absolutes with accompanying edicts. The considerations are manifold, but I'm convinced that in this modern era, we can no longer rely on a highly trained typesetting priesthood to make our publications beautiful, and it's too costly to require everyone to routinely deal with nuanced typesetting rules. So, even though double spacing between sentences makes sense in the context of fixed-width fonts (which as a programmer is what I work with the vast majority of the time), I'm no longer of the opinion that the benefits are worth the overall cost.

I've lobotomized my `rewrap` script to uniformly use single spaces between non-whitespace tokens, and will endeavor to still my twitchy right thumb.  I'll reformat this diff before committing, as well as #76.

doc/design/memory.md
```diff
@@ -0,0 +1,1595 @@
...
+
+## Data/memory model
+
+Memory is managed as pages of naturally aligned 64-bit words.  Although
+individual integral values may have sizes which are not multiples of 64 bits,
+and such values can be contiguously packed within an outer value (e.g. a `u8
+array`), the memory manager always operates on values which are whole words.
+
+Hemlock strongly favors contiguous allocation over segregated allocation; it
+copies/compacts rather than reusing freed interspersed memory.  Although some
```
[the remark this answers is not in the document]
jasone: Okay, I added a paragraph and clarified some terms.

doc/design/memory.md
jasone: Yes, that is correct.  Although we could immediately decref and deallocate a limited number of values (and we probably *will* perform one eager round of decref/deallocate), a long reference chain can trigger an arbitrarily long sequence of decref/deallocate operations.  This stack allows us to make such work incremental.

doc/design/memory.md
jasone: Suppose a space is 2 MiB.  Our page allocator isn't constrained to allocate a single 2 MiB region.  If the lazy space(s) push the resulting combined space past the 2 MiB threshold, we just allocate a multiple of the normal space size.  This same approach applies to huge allocations which don't fit in a normally sized space.

doc/design/memory.md
jasone: It will be interesting to see what utilization range ends up being practical.  I hope we can use something much higher than 0.5, but even 0.5 won't break us.

PR #79, opened 2020-07-10
Cache

jasone:


jasone:
I took a stab at improving cache hit rate, based on [this document](https://docs.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows).  TBH I have a lot of unanswered questions about how this stuff works; it feels like whoever wrote that document has a bunch of unstated background knowledge, without which the document has major holes.

jasone:
Re: the `avsm/setup-ocaml` version change, I took the advice of the [avsm docs](https://github.com/avsm/setup-ocaml/blob/master/README.md).

PR #81, opened 2020-08-03
Begin adding design documentation.

jasone:
Add the demand function (to be built in to Hemlock), which acts like assert but
cannot be disabled.


jasone:
Duplicate of #75.

PR #86, opened 2020-08-11
Scan

### LINKED ISSUE #15: Implement scanner (tokenizer)

jasone:
Resolves #15.


jasone:
The UTF-8 refactoring is worth putting in all by itself. I'll remove the `Hlc` commit momentarily, since it doesn't make sense until there's something in the new library.

PR #87, opened 2020-08-12
Array slice

cevans87:
Just a draft for now. I have `String.Slice`, `Array.Slice`, and `Bytes.Slice` all (seemingly) working. I also removed `File.Buffer` and used `Bytes.Slice` in its place.

I didn't make any effort to clean this up. Just showing it in the minimal "working" state.


jasone:
Subsumed by #89.

PR #117, opened 2021-08-22
Explicit concealed

jasone:


doc/design/effects_mutability.md
```diff
@@ -454,6 +603,32 @@ Precise effects typing of callback functions is useful for local optimization ev
...
+    val ( +* ): uns -{alloc}-> (uns --> uns)
+```
+
+As a consequence, partial application is prohibited in functions like `f`, and allowed in functions
+like `g` and `h`.
+
+```hemlock
+    val f: uns --> uns --> uns
+    val g: uns -{alloc}-> (uns --> uns)
+    val h: uns --> uns -{alloc}-> uns
```
cevans87, line 629: I don't understand how `h` works. It isn't possible to close on one parameter, line `h 1` since doing so would require a concealed `alloc` effect. If we provide two parameters, `h 1 2`, no closure is needed so the concealed `alloc` is superfluous. What am I missing?

doc/design/effects_mutability.md
```diff
@@ -215,24 +236,90 @@ type ('a, ^m, 'b, ^n, >e) ['accum, 'c] t =
...
+A function which make concealed effects explicit can only call functions which in turn transitively
+make concealed effects explicit. This restriction is obviously required for the compiler to be able
+to infer concealed effects. In practice, this means that any facility which low-level code requires
+must make its concealed effects explicit, and that such low-level code must be carefully written to
+avoid undesirable concealed effect propagation.
+
+### May-halt effect (`hlt`)
+
+The `hlt` (may-halt) effect can usually be locally handled and hidden from callers without affecting
+correctness, and is therefore concealed by default.
```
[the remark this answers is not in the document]
jasone:
```suggestion
The `hlt` (may-halt) effect can usually be hidden from callers without affecting correctness, and is
therefore concealed by default.
```

doc/design/effects_mutability.md
cevans87: Actually, I misread this. I disregarded the statement immediately above 

> However, if `+` is partially applied, the compiler transforms the function to one with an `alloc`
> effect, with signature equivalent to `+*`.
>
>```hemlock
>    val ( +* ): uns -{alloc}-> (uns --> uns)
>```

cevans87 (approved):

PR #120, opened 2021-08-30
Docker

cevans87:
Adds files and `README.md` instructions for building and running Hemlock bootstrap docker images.

See the README.md for instructions about how to build and run the docker images.

There are a few choices I made that need some justification.

-  `.dockerfile` instead of canonical `Dockerfile`
Using the canonical `Dockerifle` implies that it should be used with the `docker build` command. We explicitly want to use the `docker compose build` command to make it easier to set up environment variables, volumes, and options when we run the container.

- `.docker-compose.yml` instead of canonical `docker-compose.yml`
This one is simply to make the root of our repo look cleaner when typing `ls`. There is no other reason. `docker compose` detects the non-canonically-named file via the `COMPOSE_FILE` variable in the `.env` file at the root of the Hemlock repo.


jasone (commented):
Can you please re-organize this to be two commits, where the first is `Reformat README.md to 100 columns`, and the second is everything else squashed? It's probably ready to go otherwise.

README.md
```diff
@@ -14,6 +14,56 @@ not yet suited for any sort of production use.
 
 ### Prerequisites
 
+There are two separate methods for building the bootstrap compiler (Docker and Native). Follow
+prerequisites for your preferred method.
+
+#### (Preferred) Docker Prerequisites
```
jasone, line 20: `(Preferred)` here is confusing. I first thought it meant that these are the preferred docker-specific instructions, but that there's an alternative docker approach.

jasone:
```suggestion
#### Docker Prerequisites
```

jasone (approved):

PR #122, opened 2021-09-03
Fix ! and $ scanning

jasone:
This diff is incidentally based on a couple of tooling updates.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,7 +21,7 @@ jobs:
     - name: opam
       uses: avsm/setup-ocaml@v1
       with:
-        ocaml-version: '4.11.1'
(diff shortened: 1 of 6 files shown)


.github/workflows/test.yml
```diff
@@ -21,7 +21,7 @@ jobs:
     - name: opam
       uses: avsm/setup-ocaml@v1
       with:
-        ocaml-version: '4.11.1'
+        ocaml-version: '4.12.0'
```
cevans87, line 24: I think there's a way to pull this from the top-level `.env` file, but figuring out how to do that probably isn't important since we want to switch to using our `prod` docker image, which already uses that file.

cevans87 (approved):

PR #123, opened 2021-09-09
Enable multi-platform Docker build.

cevans87:
This enables us to manually set the target platform for our image using the HEMLOCK_PLATFORM
environment variable. There are less-verbose ways of plumbing the target platform through, such as
using the `platform: ` key in docker-compose.yml. However, by doing it in here, the default value
for HEMLOCK_PLATFORM is set equal to the docker buildx BUILDPLATFORM environment variable, which
matches the builder's host platform. That means that in the cases where we connect a remote builder
(AWS Graviton, for instance), the default target platform will match that of the builder and not of
the local host where the CLI commands are invoked.

### DIFF
--- a/.docker-compose.yml
+++ b/.docker-compose.yml
@@ -4,6 +4,7 @@ services:
     build:
       args:
         - HEMLOCK_BOOTSTRAP_OCAML_VERSION
+        - HEMLOCK_PLATFORM
       context: .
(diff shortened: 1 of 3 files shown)


jasone:
The CI failures wouldn't have happened if we were using Docker images there. :wink:

jasone (approved):
Nice!

PR #124, opened 2021-09-12
Rename 'master' branch to 'main'.

jasone:
Default branch naming inconsistency among repos is starting to chafe.

CI is going to fail for this commit, since the branch rename cannot be done as part of the commit itself.

### DIFF
--- a/.github/scripts/check_ocp_indent.py
+++ b/.github/scripts/check_ocp_indent.py
@@ -60,11 +60,11 @@ class Hunk:
 
     @staticmethod
     async def from_path(path: Path, /) -> Tuple[Hunk]:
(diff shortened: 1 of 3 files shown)


jasone:
The branch rename operation is via web interface. We'll need to do this for every cloned local repo:

```sh
 git branch -m master main
 git fetch origin
 git branch -u origin/main main
 git remote set-head origin -a
 git remote prune origin
```

A similar dance is required for each forked repo. I've transitioned every other repo I have control over, but this one will be a bit more disruptive since we both have feature branches in progress. That said, feature branches don't prevent successful renaming and clean end results.

cevans87 (approved):

PR #126, opened 2021-09-13
Set default dockerfile ubuntu tag to "rolling".

cevans87:
The default Ubuntu image will now follow the rolling releases of Ubuntu. Our images were previously
based on the "latest" tag, which follows the LTS release. You can set the "HEMLOCK_UBUNTU_TAG" env
variable to another valid ubuntu image tag if you need to test against another release. See
https://hub.docker.com/_/ubuntu?tab=description&page=1&ordering=last_updated


cevans87:
I got annoyed that I had meson version .53. I want the newer one.

jasone:
Cool. I recently hit a ruby bundler version incompatibility between host and container that this will also address.

jasone (approved):

PR #130, opened 2021-09-17
Existentials

jasone:
See individual commit messages for an overview of changes. I'm not entirely sure I like the effect declaration syntax, but it's better than the previous ~5 I tried; suggestions for improvement welcome.

### DIFF
--- a/doc/design/effects_mutability.md
+++ b/doc/design/effects_mutability.md
@@ -32,59 +32,54 @@ made between operating system load versus store effects.
 
 For more semantics detail than is provided in the following syntax quick reference, see below.
 
-- Effects
-  - `>{>e,>f}`, `>e`, `>`: Parametric effect(s). The `>e` and `>` short forms can be used only if
-    there is a single parametric effect.
-  - `>_`: Effects are contextually irrelevant, and therefore arbitrarily permitted.
+- Effect (`effect`)
+  - + `e`, `{e}` `{e|f}`, `{e|os|hlt}`: Effect sets. A set with only one member may omit the `{...}`
+    enclosure. Empty effect sets have no explicit syntax.
   - Select pre-defined effects. See [Effects categories](#effects-categories) for more.
-    + `>{os}`: Operating system load-store effect
-    + `>{ld}`: Runtime system load effect
-    + `>{mut}`: Runtime system load-store effect
-    + `>{hlt}`: Runtime system may-halt effect
-  - `>{os,hlt}`: Set of effects.
-  - `>{>e,>f}\{os,mut}`, `>\{mut}`: Effects constraint; transitively prohibited effects are
-    enumerated as `\{...}` following any effects.
-  - `('a t&! -> unit)`, `('a&! t& >{hlt}-> unit)`: Mutation effect on mutable function
-    parameter type suffixed by `!`. The affected type need not be the outermost type, but the
-    affected type must be mutable.
-  - `('a t& -> 'a t)`: Load effect on mutable function parameter. A mutable function parameter
-    always implies a `ld` effect on the parameter, without which the parameter would be useless.
-  - `(uns >{t&,u&!}-> unit)`, `(t&! >{t&!}-> unit)` : Type-constrained environment `ld`/`mut` effect
-    during function application, listed in `>{...}->`. Note that a `t&!` parameter mutation is
-    distinct from `>{t&!}->` environment mutation.
-  - `(uns >{ld}-> unit)`, `(uns >{mut}-> unit)`: Environment `ld`/`mut` effect during function
-    application. Type-agnostic environment effects impact mutable values of all types.
-  - `-{alloc}`, `-{alloc,hlt}`: Explicit [never-parametric] concealed effect(s). An explicitly pure
-    application is written as `-->`, which is shorthand for `-{}->` (or more verbosely,
-    `>{}\{}-{}->`).
-- Mutability
-  - `^m t`, `^m t ^t`, `^m t^`: Parametric mutability. `^m` is an subsidiary mutability parameter,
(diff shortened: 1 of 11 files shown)


jasone:
I realized there were some problems with allowing opaque effect declarations, and simplified the syntax while fixing the semantic issues. I also went to `{...|...}` effect set syntax, because the `expose`/`conceal` syntax was really confusing in practice. This reverts us to `{}`-delimited effect sets, but using `|` as the member separator rather than `,`, so I think this resolves the ambiguity issue we were trying to resolve with the `>e>f->` syntax.

doc/design/effects_mutability.md
```diff
@@ -32,59 +32,54 @@ made between operating system load versus store effects.
...
+  - `val v ^m: t m`, `val v ^t ^m: ^t&t m`, `val v ^t ^m: ^&t m`: Parametric mutability. `^m` is a
+    subsidiary mutability parameter, whereas `^t` comes after the type name and is the primary
+    mutability parameter. `^t&t` can be shortened to `^&t`.
+  - `_&t`: Mutability is contextually irrelevant, and therefore both immutable and mutable values
+    are supported.
+  - `&t`: Non-parametric primary mutability. `!&t` indicates a mutation effect on the value, which
+    must therefore be mutable, as in `(a: type) -> ... -> !&t a -> ...`.
+  - `\&t`: Transitive prohibited mutability constraint.
+- Type (`type`)
+  - `type a 'a: a`, `type a 'aa ^a ^m >e: a^ aa m e`: Parametric type.
```
cevans87, line 70: It looks like `a^` is a shortening of `a^a`. It's a little weird that shortened syntax differs from `^t&t` to `^&t`, though I see why `a^a` can't be shortened to `^a` using a similar rule (it's ambiguous with the effect parameter syntax).

It makes me wonder if it would be better to just apply the shortening sigil on the right side all the time to keep things consistent.
`^t&t` shortens to `&t^` instead of `^&t`. Perhaps it'd be better to write `^t&t` as `&t^t` and say `&t^t` shortens to `&t^`. It also fits a little better in my head to attach the primary mutability parameter on the right side of the type (almost like a parameter to the type, except without a space to indicate primary mutability).

doc/design/effects_mutability.md
```diff
@@ -114,82 +109,68 @@ That said, although interfaces are conservatively inferred, the compiler is free
...
 
 ```hemlock
-'a list
-'a array^
-('k\&, 'v, 'cmp) map
-('a\&, 'cmp) set
+type list 'a: 'list a
+type array 'a ^array: ^&array a
+type map 'k 'v 'cmp: map \&k v cmp
+type set 'a 'cmu: set \&a cmp
```
cevans87, line 149:
```suggestion
type set 'a 'cmp: set \&a cmp
```

doc/design/effects_mutability.md
```diff
@@ -224,85 +213,61 @@ arises outside of function signatures. Even function signatures rarely require t
 parametric effects.
 
 ```hemlock
-val unzip_map: ('a >e-> 'c) -> ('b >f-> 'd) -> ('a * 'b) list
-  >{>e,>f}-> 'c list * 'd list
+val unzip_map 'a 'b 'c 'd >e >f: (a >e-> c) -> (b >f-> d) -> (a, b) list >{e|f}-> (list c, list d)
```
cevans87, line 216:
```suggestion
val unzip_map 'a 'b 'c 'd >e >f: (a >e-> c) -> (b >f-> d) -> list (a, b) >{e|f}-> (list c, list d)
```

cevans87 (commented):
Leaving these comments here for discussion. I'm still going through `array.hm` and `array.hmi`.

doc/design/syntax.md
```diff
@@ -115,10 +115,10 @@ The following words are keywords which are used as syntactic elements, and canno
...
 
 ```hemlock
-and      external  match   val
-also     false     module  when
-as       for       of      while
-assert   fun       open    with
+and      external  match       val
+also     false     mutability  when
+as       for       of          while
+assert   fun       open        with
```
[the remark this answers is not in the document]
jasone: This is actually a really annoying table, because Markdown doesn't provide any sort of tabular output that keeps the keyword list from being ridiculous. So it's manually formatted.

doc/design/effects_mutability.md
```diff
@@ -32,59 +32,54 @@ made between operating system load versus store effects.
...
+    the affected type must be mutable.
+  - `(a: type) -> &t a -> &t a`: Load effect on mutable function parameter. A mutable function
+    parameter always implies a `ld` effect on the parameter, without which the parameter would be
+    useless.
+  - `uns >{&t|(!&u)}-> unit`, `!&t >!&t-> unit`: Type-constrained environment `ld`/`mut` effect(s)
+    during function application. Parentheses around type-constrained effects are optional, though
+    they often improve readability. Note that a `!&t` parameter mutation is distinct from
+    `>(!&t)->` environment mutation.
+  - `uns >ld-> unit`, `uns >mut-> unit`: Environment `ld`/`mut` effect during function application.
+    Type-agnostic environment effects impact mutable values of all types.
```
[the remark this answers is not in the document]
jasone: As discussed offline: These examples show how a function can mutate values that were captured in a lexical closure. Often the compiler can be more accurate than `ld` or `mut`, but if the mutation is of a module field binding and the lexical closure was from within the module, the type cannot be more specific because that would require recursive module types.

doc/design/effects_mutability.md
jasone: Whoops, `a^` should be `^&a`.

cevans87 (approved):

PR #133, opened 2021-09-19
Rewrap OCaml and C code to 100 columns.

jasone:
This was a very manual effort, so no guarantees I caught everything, but it should be a way better starting point than incrementally rewrapping every time we modify existing code.


cevans87:
Blindly stamping this one so I can use it. Let me know if you actually need a more thorough review.

cevans87 (approved):

PR #134, opened 2021-09-20
Remove .merlin file.

jasone:
Merlin configuration has been auto-generated by dune for several releases, and
the presence of a .merlin file prevents proper function of ocaml-lsp.


cevans87:
Fixes things for me too.

cevans87 (approved):

PR #135, opened 2021-09-20
Fix missing '\0' termination in deflate_bytes().

### LINKED ISSUE #129: Inconsistent '\0' termination in hemlock_file_of_path_inner()
This command fails:
```sh
hemlock@f80531087964:~/Hemlock/bootstrap [main]> dune exec hmc /home/hemlock/Hemlock/src/Basis/Array.hm                                                                                                                                                                                                                                                                                                                                           Sep 16 17:32:45
Halt: File.of_path error: No such file or directory
```
However, this works fine:
```sh
hemlock@f80531087964:~/Hemlock/bootstrap [1ml_inspired_syntax]> dune exec hmc /home/hemlock/Hemlock/src/Basis/Array.hmi
```
This appears to be due to a length-dependent `'\0'` termination bug in `hemlock_file_of_path_inner()`. The following diff makes the error go away, but I didn't investigate in enough detail to know whether this is a reasonable fix.

```diff
diff --git a/bootstrap/src/basis/file.c b/bootstrap/src/basis/file.c
index dc1ce73..dde1483 100644
--- a/bootstrap/src/basis/file.c
+++ b/bootstrap/src/basis/file.c
@@ -84,6 +84,7 @@ hemlock_file_of_path_inner(value a_flag, value a_mode, value a_i, value a_j,
   size_t n = j - i;
   bytes = &bytes[i];
   uint8_t * cbytes = deflate_bytes(n, bytes);
+  cbytes[n] = '\0';
   // TODO expose mode in API
   int result = open(cbytes, flags, mode);
   inflate_bytes(n, cbytes);
```

cevans87:
Fixes #129

Manual test with the following at the command line.
```
dune exec hmc /home/hemlock/Hemlock/src/Basis/Array.hm
```
Before the change, this results in the output
```
Halt: File.of_path error: No such file or directory
```
Array.hm clearly exists at this location in the Hemlock image, so this was a bug.

After the changes in this PR, the result is that the test print the expected `hmc` output.

`deflate_bytes()` is only used to prepare a `Bytes.t` to be used with a system call. A `Bytes.t` in
memory actually uses 32 bits per byte, so a `Bytes.t` with the value `"path"` might look like
```
? ? ? p ? ? ? a ? ? ? t ? ? ? h
```
After `deflate_bytes()`, it looks like
```
p a t h ? ? ? a ? ? ? t ? ? ? h
```
To be safe with a system call, it needs to be null-terminated.
```
p a t h\0 ? ? a ? ? ? t ? ? ? h
```

The `deflate_bytes()` conterpart, `inflate_bytes()` does not have a similar bug.

### DIFF
--- a/bootstrap/src/basis/file.c
+++ b/bootstrap/src/basis/file.c
@@ -28,6 +28,9 @@ deflate_bytes(size_t n, size_t * inflated_bytes) {
     for (size_t i = 0; i < n; i++) {
         deflated_bytes[i] = Int_val(inflated_bytes[i]);
     }
+    if (n > 0) {
+        deflated_bytes[n] = '\0';
+    }
     return deflated_bytes;
 }
 


bootstrap/src/basis/file.c
```diff
@@ -28,6 +28,9 @@ deflate_bytes(size_t n, size_t * inflated_bytes) {
     for (size_t i = 0; i < n; i++) {
         deflated_bytes[i] = Int_val(inflated_bytes[i]);
     }
+    if (n > 0) {
```
jasone, line 31: :+1:

jasone (approved):

PR #136, opened 2021-09-23
No ppx

jasone:


cevans87 (approved):
Made it through about 1/10 of the new test files. It's nice to have something closer to what hemlock will look like.

With the new dune files listing off all the associated test files, is there a way to be certain we caught them all?

jasone:
Unfortunately, I don't think there is an automated way to detect tests that are accidentally omitted in the corresponding `dune` files. The tests get built as modules no matter what, but they don't get executed as test unless explicitly listed.

PR #137, opened 2021-09-23
Update verbiage to refer to seven (not nine) regions per actor.

jasone:
The corresponding figure correctly depicts seven regions.


cevans87 (approved):
I remember that being a puzzling number.

PR #139, opened 2021-09-29
Convert uns/sint from int-based to int64-based.

### LINKED ISSUE #100: Implement `uns`/`sint` using `int64` rather than `int`
The fact that `uns` is actually `u63` rather than `u64` is causing edge case handling friction in the bootstrap compiler. Refactor the bootstrap `Basis` library to use `int64` rather than `int` as the basis for `uns` and `sint`. This will require using e.g. `0L` rather than `0` for `uns`/`sint` literals, but OCaml _does_ have built-in syntactic support for `int64`, including for pattern matching. Performance will suffer some due to OCaml boxing `int64` but the hit will be acceptable for bootstrapping purposes.

jasone:
Resolves #100.


bootstrap/src/basis/file.c
```diff
@@ -7,6 +7,7 @@
 #include <unistd.h>
 
 #include <caml/mlvalues.h>
+#include <caml/alloc.h>
```
cevans87, line 10: I wish I'd thought to look for this while writing this file. Some of the functions in here dealing with arrays could have been much simpler.

cevans87 (approved):

Issue #143, opened 2021-09-29
Add ranges


jasone:
Refactor `Basis` to use ranges as the foundation for integer-based iteration, as well as for slices. Right now iteration over a range of integers is ad hoc, in particular because OCaml's `for...to...done` construct can only operate on `int` and `Basis` now exclusively uses `int64` to implement its integer types. Furthermore, slices comprise [`base`..`past`) cursors, which is an unfortunate representation because both cursors refer to the container.

The refactor for the bootstrap `Basis` will require roughly the following steps:
- [ ] Expand `ContainerCommonIntf.S{Mono,Poly}Fold` to have:
  + [ ] `foldi_right_until`
  + [ ] `foldi_right`
  + [ ] `iter_right`
  + [ ] `iteri_right`
- [ ] Add range interfaces which extend `ContainerIntf.SMono`:
  + [ ] `RangeIntf` (half-open) has a `=:<` initialization function (will be `..` in Hemlock)
  + [ ] `RangeFIntf` (full-open) has a `=:=` initialization function (will be `..=` in Hemlock)
- [ ] Add `Range` and `RangeF` modules to `Intnb`. Implement using [`base`..`past`) and [`left`..=`right`] conventions.
- [ ] Open `Uns.Range` in `RudimentsInt`.
- [ ] Refactor slices to store {container, range} instead of {base, past}.
- [ ] Expand `SliceIntf.S{Mono,Poly}` to have:
  + [ ] `?range` parameter for `of_container`
  + [ ] `range`
- [ ] Refactor tests' `iter_oc` to use `Uns.Range.iter` (typically as `iter (0 .. 5) ~f:(fun i -> ... )`).

jasone:
Actually, the `*i_right*` functions don't make a lot of sense. We don't know the length of containers prior to completing iteration, so the only index we can provide during iteration is relative to the end of the container. No matter, `iter_right` provides the only thing we're missing with regard to folding/iterating over ranges.

jasone:
Storing {container, range} instead of {base, past} in slices doesn't actually work, because slices don't presume cursors can be created at arbitrary offsets in constant time (e.g. string cursors require O(n) iteration depending on indexing mode).

Issue #99, opened 2021-01-10
Implement comprehensive integer conversion function matrix


jasone:

jasone:
The conversion matrix has to be implemented differently than described, in order to avoid cyclic dependencies among types. The approach I'm taking is to order the type dependencies as `u8` -> `i8` -> `u16` -> `i16` -> ... -> `u512` -> `i512` -> `nat` -> `zint`. Furthermore, the conversions to/from e.g. `u8` and `i128` are all in the `U8` module.

jasone:
I have the `[ui]{128,256,512}` conversions implemented, but there's a ridiculous amount of boilerplate. I'm going to functorize the code before finishing off the conversion matrix so that the boilerplate doesn't get completely out of hand.

jasone:
Still plodding along on this. All the conversion functors are now implemented, but `Uns`/`Sint` can't use them to convert to/from `Intw`-based types due to cyclical dependencies. I'm part way through breaking the dependency loop.

PR #144, opened 2021-09-30
Range

jasone:


jasone:
This is coming along, though it's going to take at least another couple days to finish up, because #99 turns out to be a prerequisite for #143.

jasone:
This is feature-complete now, but testing is spotty.

jasone:
All ready now (AFAIK).

jasone:
I converted the last few straggler `for` loops to range iteration, and a a side effect discovered a couple tests which weren't enumerated in `dune` files. That prompted me to write a script to search for other disconnected tests, and there are none remaining.

PR #170, opened 2021-11-09
Refactor File module to use io_uring

cevans87:


bootstrap/src/basis/file.ml
```diff
@@ -201,38 +198,41 @@ let read_into_hlt buffer t =
   let _ = Read.(submit ~buffer t |> complete_hlt) in
   ()
 
-(* external write_inner: _?bytes -> t >os-> sint *)
-external write_inner: Stdlib.Bytes.t -> t -> sint = "hm_basis_file_write_inner"
-
-let rec write buffer t =
```
cevans87, line 142: We recursively called write until the entire buffer was written here. We don't handle the same problem in the new version. Make sure there isn't a subtle bug here.

cevans87: This was an actual problem. Fixed now.

bootstrap/src/basis/ioring.c
```diff
@@ -0,0 +1,554 @@
...
+    sqe->off = -1;  // Use current file position.
+    sqe->fd = fd;
+    sqe->addr = (uint64_t)buffer;
+    sqe->len = n;
+
+OUT:
+    assert(oe == HM_OE_NONE);
+
+    return user_data;
+}
```
cevans87, line 530: I'm not it's okay to just swallow a possible error in the C layer. We need to return a pointer (as `uns`) to OCaml so there's no way to return errors as negative values like in some of the other OCaml/C bindings.

That'd mean I'd need to return a `(sint * uns)` from C for the error and the user_data pointer. I'd also have to handle the error on the OCaml side. APIs like `File.Read.submit` would change slightly.

bootstrap/src/basis/file.ml
```diff
@@ -130,49 +130,76 @@ let close t =
...
 
 let read_into_hlt buffer t =
-  match read_into buffer t with
-  | None -> ()
-  | Some error -> begin
-      let _ = close_hlt t in
-      halt (Error.to_string error)
-    end
+  let _ = Read.(submit ~buffer t |> complete_hlt) in
+  ()
```
cevans87, line 194: I think the `read_into` functions are somewhat misguided. I'd originally convinced myself they were needed so that Hemlock would see the `read` function as having no mutation effects and `read_into` having mutation effects. If we were to just have a single `read` function that took an optional input buffer, we could mutate an input, but I don't think that's such a bad problem to warrant these additional functions.

bootstrap/src/basis/ioring.c
cevans87: I made the change I suggested here and will push it along with fixes from the current round of review.

bootstrap/src/basis/file.mli
```diff
@@ -39,6 +39,25 @@ val stdout: t
...
+  (** [submit ~flag ~mode path] submits an open operation for a file at [path] with [flag] (default
+      Flag.R_O) Unix file permissions and [mode] (default 0o660) Unix file permissions. This
+      operation does not block. *)
+
+  val complete: t -> (file, Error.t) result
+  (** [complete t] blocks until give [t] is complete. Returns a file or an error if the file could
+      not be opened. *)
+
+  val complete_hlt: t -> file
+  (** [complete_hlt t] blocks until give [t] is complete. Returns a file or halts if the file could
```
jasone, line 50: s/give [t]/[t]/g ?

bootstrap/src/basis/file.mli
```diff
@@ -70,6 +70,33 @@ val of_path_hlt: ?flag:Flag.t -> ?mode:uns -> Bytes.Slice.t -> t
...
+module Close: sig
+  type file = t
+  type t
+  (* An internally immutable token backed by an external I/O close completion data structure. *)
+
+  val submit: file -> t
+  (** [submit file] submits a close for given [file]. This operation does not block. *)
+
+  val complete: t -> Error.t option
+  (** [complete t] blocks until the given [t] is complete. Returns None or an error if file could
```
jasone, line 73: s/None/[None]/ (same with other language literals like `unit`, `Some error`, etc.)

bootstrap/src/basis/file.ml
```diff
@@ -130,49 +130,76 @@ let close t =
...
+  let complete t =
+    let base = Bytes.(Cursor.index (Slice.base t.buffer)) in
+    let bytes = Stdlib.Bytes.create (Int64.to_int (Bytes.Slice.length t.buffer)) in
+    let value = complete_inner bytes t.inner in
+    match Sint.(value < kv 0L) with
+    | true -> Error (Uns.bits_of_sint Sint.(neg value))
+    | false -> begin
+        let range = (base =:< (base + (Uns.bits_of_sint value))) in
+        let container = Bytes.Slice.container t.buffer in
+        Range.iter range ~f:(fun i ->
```
jasone, line 167: I think `Range.iter` can just be `iter` here.

bootstrap/src/basis/file.ml
```diff
@@ -201,38 +198,67 @@ let read_into_hlt buffer t =
...
+    inner: inner;
+    buffer: Bytes.Slice.t;
+  }
+
+  external submit_inner: Stdlib.Bytes.t -> file -> inner = "hm_basis_file_write_submit_inner"
+
+  let submit buffer file =
+    let bytes = bytes_of_slice buffer in
+    let inner = finalize_user_data (submit_inner bytes file) in
+    { inner; buffer }
```
jasone, line 209: `{inner; buffer}`

bootstrap/src/basis/file.ml
```diff
@@ -201,38 +198,67 @@ let read_into_hlt buffer t =
...
+        let past = Bytes.Slice.past t.buffer in
+        let buffer = Bytes.Slice.of_cursors ~base ~past in
+        Ok buffer
+      end
+
+  let complete_hlt t =
+    match complete t with
+    | Ok buffer -> buffer
+    | Error error -> halt (Error.to_string error)
+
```
jasone, line 226: Extraneous whitespace.

bootstrap/src/basis/file.ml
```diff
@@ -201,38 +198,67 @@ let read_into_hlt buffer t =
...
+end
+
+let write buffer t =
+  let rec f buffer t = begin
+    match Bytes.Slice.length buffer = 0L with
```
jasone, line 231: Maybe put an `assert` in the C layer that length is 0.

jasone (commented):
Overall this is super clean, very nice. My high-level thoughts are:

- Comments would make some of this code easier to understand/maintain. In particular, queue data structures and semantics could do with a brief high-level description.
- The pretty printers are too complicated, and IMO the output is actually worse off for it.

bootstrap/src/basis/file.ml
cevans87: Do you mean put it there so that we don't have to recursively call `write` here? As far as I can tell from reading `man 2 write` `RETURN VALUE` section, partial writes shouldn't be considered an error.

jasone: No, I mean that this code assures the C layer never gets used for 0-length writes, and that we should `assert` in the C code that the write length is non-zero. That documents expectations in the C layer, so that if we ever refactor in a way that makes this no longer true we are at least made aware enough to add extra tests.

bootstrap/src/basis/ioring.c
```diff
@@ -0,0 +1,462 @@
...
+    return syscall(__NR_io_uring_setup, entries, p);
+}
+
+int io_uring_enter(unsigned int fd, uint32_t to_submit, uint32_t min_complete, uint32_t flags) {
+    return syscall(__NR_io_uring_enter, fd, to_submit, min_complete, flags, (size_t) NULL);
+}
+
+void hm_array_pp(int fd, int indent, unsigned len, unsigned *array) {
+    bool well_formed = true;
+    for (size_t i = 0; i < len && well_formed; i++) {
```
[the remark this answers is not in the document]
cevans87: `unsigned` is the type used in the io_uring API. Isn't it 32 bits? I meant to use it appropriately for io_uring data structure access. `unsigned *array` is intentional. You're right that the `unsigned len` is a mistake.

bootstrap/src/basis/executor.c
```diff
@@ -0,0 +1,23 @@
+#include <threads.h>
+
+#include "common.h"
+#include "ioring.h"
+
+#include "executor.h"
+
+thread_local hm_executor_t executor = {0};
```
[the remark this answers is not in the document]
cevans87: Yeah. I had to dig into this one a bit because I'd originally written the `{}` as `{0}` and it bit me hard. This is a shorthand for zeroing the first field. All remaining fields take the same value.

bootstrap/src/basis/file.c
```diff
@@ -132,3 +137,79 @@ hm_basis_file_seek_tl_inner(value a_i, value a_fd) {
...
+
+    return Val_unit;
+}
+
+// val hm_basis_file_user_data_pp: Basis.File.{Open|Close|Read|Write}.t >{os}-> unit
+CAMLprim value
+hm_basis_file_user_data_pp(value a_user_data) {
+    hm_user_data_t *user_data = (hm_user_data_t *)Int64_val(a_user_data);
+
+    hm_user_data_pp(STDOUT_FILENO, 0, user_data);
```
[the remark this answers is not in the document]
cevans87: I tested changing these pretty-printers to use `stderr` and `dune runtest` failed to capture the output. Am I missing something?

bootstrap/src/basis/ioring.c
```diff
@@ -0,0 +1,462 @@
...
+#include <string.h>
+#include <stdlib.h>
+#include <sys/mman.h>
+#include <sys/syscall.h>
+#include <sys/types.h>
+#include <unistd.h>
+
+#include "common.h"
+#include "ioring.h"
+
```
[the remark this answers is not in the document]
cevans87: Ugh, habit from Python.

bootstrap/src/basis/file.c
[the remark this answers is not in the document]
jasone: I want these pretty printers to be usable when I'm debugging a program which is writing non-debug output to `stdout`. The simplest approach is typically to write to `stderr`. I understand that the tests need the pretty printing output to go to `stdout`, but with that hard-coded in the APIs the pretty printers' general utility is significantly reduced.

bootstrap/src/basis/ioring.c
```diff
@@ -0,0 +1,462 @@
...
+            dprintf(fd, ", %2u", array[i]);
+        }
+        dprintf(fd, " ]\n");
+    }
+}
+
+void hm_opcode_pp(int fd, int indent, unsigned opcode) {
+    switch (opcode) {
+    case IORING_OP_NOP:
+      dprintf(fd, "%*sopcode:   IORING_OP_NOP\n", indent, "");
```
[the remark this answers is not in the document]
cevans87: I think it made it vertically align with other values. I'm getting rid of the crazy formatting logic, so the three spaces are going away.

bootstrap/src/basis/ioring.c
```diff
@@ -0,0 +1,462 @@
...
+hm_ioring_flush_cqes(uint32_t min_complete, hm_ioring_t *ioring) {
+    hm_opt_error_t oe = HM_OE_NONE;
+    hm_cqring_t *cqring = &ioring->cqring;
+
+    assert(min_complete <= *cqring->ring_entries);
+
+    uint32_t tail = *cqring->tail;
+    uint32_t head = HM_SMP_LOAD_ACQUIRE(cqring->head);
+    if (tail - head < min_complete) {
+        uint32_t n_complete_;
```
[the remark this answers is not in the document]
cevans87: I think it was to avoid a name conflict when I still had the `min_complete` input parameter named `n_complete`.

bootstrap/src/basis/ioring.c
```diff
@@ -0,0 +1,462 @@
...
+hm_ioring_get_sqes_size(hm_ioring_t *ioring) {
+    return ioring->params.sq_entries * sizeof(struct io_uring_sqe);
+}
+
+hm_opt_error_t
+hm_ioring_setup(hm_ioring_t *ioring) {
+    hm_opt_error_t oe = HM_OE_NONE;
+
+    memset(ioring, 0, sizeof(*ioring));
+    HM_OE_ERRNO_RESULT(oe, ioring->fd, io_uring_setup(32, &ioring->params));
```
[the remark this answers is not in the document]
cevans87: It's the default used in examples, and the value needs to be a power of 2. I'll comment it.

bootstrap/test/basis/file/test_file_full_sq.expected
```diff
@@ -0,0 +1,420 @@
+sqring:
```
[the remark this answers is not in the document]
cevans87: The choice of 2 vs 4 seemed arbitrary. As far as making this logic easier to deal with, flattening the entire pretty-printer output into a single-line json blob would be my preference.

bootstrap/src/basis/ioring.h
```diff
@@ -23,32 +35,42 @@ typedef struct {
...
     unsigned *ring_entries;
     unsigned *ring_mask;
     struct io_uring_cqe *cqes;
 } hm_cqring_t;
-void hm_cqring_pp(int fd, int indent, hm_cqring_t *cqring);
+void hm_cqring_pp(int fd, hm_cqring_t *cqring);
 
+// Utility type for tracking io_uring fd and mmapped data structure fields.
 typedef struct {
+    // Parameters for co
```
cevans87, line 52: Incomplete comment.

bootstrap/src/basis/file.ml
```diff
@@ -61,20 +61,29 @@ let stderr =
 external user_data_decref: uns -> unit = "hm_basis_file_user_data_decref"
 external complete_inner: uns -> sint = "hm_basis_file_complete_inner"
 
-let finalize_user_data user_data =
+let register_user_data_finalizer user_data =
   let () = Stdlib.Gc.finalise user_data_decref user_data in
```
cevans87, line 65: There's a possibility that `user_data` comes back uninitialized as a zero. Check for that here.

bootstrap/src/basis/ioring.c
```diff
@@ -265,86 +238,84 @@ hm_cqring_setup(void *vm, struct io_cqring_offsets const *offsets, hm_cqring_t *
...
-hm_opt_error_t
+void
 hm_ioring_teardown(hm_ioring_t *ioring) {
-    hm_opt_error_t oe = HM_OE_NONE;
-
-    HM_OE_ERRNO_OPTION(oe, munmap(ioring->sqring.sqes, hm_ioring_get_sqes_size(ioring)));
-    HM_OE_ERRNO_OPTION(oe, munmap(ioring->vm, hm_ioring_get_vm_size(ioring)));
-    HM_OE_ERRNO_OPTION(oe, close(ioring->fd));
+    if (munmap(ioring->sqring.sqes, hm_ioring_get_sqes_size(ioring)) != 0 ||
+      munmap(ioring->vm, hm_ioring_get_vm_size(ioring)) !=0 ||
```
cevans87, line 296: Missing space between `!=` and `0`.

bootstrap/src/basis/file.c
```diff
@@ -53,82 +60,242 @@ hm_basis_file_finalize_result(int result) {
...
+    hm_user_data_t *user_data = (hm_user_data_t *)Int64_val(a_user_data);
     int fd = Int64_val(a_fd);
 
-    return hm_basis_file_finalize_result(close(fd));
+    hm_user_data_pp(fd, user_data);
+
+    return Val_unit;
+}
+
+// hm_basis_file_complete_inner: Basis.File.{Open|Close|Write}.t >{os}-> int
```
cevans87, line 132: It occurred to me that this signature can't be correct. If `complete` simply had an `>{os}->` effect, it'd be valid to package it up in a lazy closure and stuff it onto the global heap, allowing any other actor to perform the completion. That's bad, because a submission sets up space within the associated actor for the completion result. I was a little worried that this implied we might need something like a `>{complete}->` effect to prevent this, but I think the answer might be to just make `Basis.File.{Open|Close|Write}.t` mutable. It'd make it valid for the mutable types to exist in an immutable lazy closure only if it was created and completed within a single `Lazy.force`. All that would be done within the same actor.

This isn't anything new. We discussed it in the past. I just realized I'd gotten these signature comments wrong.

bootstrap/test/basis/file/test_file_full_sq.expected
```diff
@@ -0,0 +1,3 @@
...
...ags": 0, "ioprio": 0, "fd": 0, "opcode": "IORING_OP_NOP", "user_data": {"refcount": 1, "opcode": "IORING_OP_NOP", "cqe": {"res": 0, "flags": 0, "user_data": "NULL"}}}, {"flags": 0, "ioprio": 0, "fd": 0, "opcode": "IORING_OP_NOP", "user_data": {"refcount": 1, "opcode": "IORING_OP_NOP", "cqe": {"res": 0, "flags": 0, "user_data": "NULL"}}}, {"flags": 0, "ioprio": 0, "fd": 0, "opcode": "IORING_OP_NOP", "user_data": {"refcount": 1, "opcode": "IORING_OP_NOP", "cqe": {"res": 0, "flags": 0, "user_data": "NULL"}}}, {"flags": 0, "ioprio": 0, "fd": 0, "opcode": "IORING_OP_NOP", "user_data": {"refcount": 1, "opcode": "IORING_OP_NOP", "cqe": {"res": 0, "flags": 0, "user_data": "NULL"}}}, {"flags": 0, "ioprio": 0, "fd": 0, "opcode": "IORING_OP_NOP", "user_data": {"refcount": 1, "opcode": "IORING_OP_NOP", "cqe": {"res": 0, "flags": 0, "user_data": "NULL"}}}, {"flags": 0, "ioprio": 0, "fd": 0, "opcode": "IORING_OP_NOP", "user_data": {"refcount": 1, "opcode": "IORING_OP_NOP", "cqe": {"res": 0, "flags": 0, "user_data": "NULL"}}}, {"flags": 0, "ioprio": 0, "fd": 0, "opcode": "IORING_OP_NOP", "user_data": {"refcount": 1, "opcode": "IORING_OP_NOP", "cqe": {"res": 0, "flags": 0, "user_data": "NULL"}}}]}
```
jasone, line 2: 😢 Might as well just print a hash value.

jasone (changes requested):
I started going through this, but found there are at least several comments from my previous review you didn't address, so I didn't dig in as deeply as intended. Can you please make another pass?

bootstrap/src/basis/file.ml
cevans87: Tried it. Doesn't seem to work.

bootstrap/src/basis/common.h
```diff
@@ -0,0 +1,27 @@
...
+            goto LABEL_OUT; \
+        }; \
+    } while (0)
+
+#define HM_OE_ERRNO_RESULT(oe, result, statement) do { \
+        result = statement; \
+        if ((int64_t)result == -1) { \
+            HM_OE(oe, errno); \
+        }; \
+    } while (0)
```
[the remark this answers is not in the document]
cevans87: I'm not sure what needs to be fixed here. My previous version had unintentional 2-space indents, but I thought what I have here is standard.

bootstrap/test/basis/file/test_file_full_sq.expected
cevans87: I'm going to think about this a little. We have a convenient string hash in our Basis library, but it might take a bit of effort (or ugly hackery) to get this output into OCaml. Do you think this is spammy enough that we need to go through the effort?

bootstrap/src/basis/common.h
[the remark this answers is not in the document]
cevans87:
```suggestion
#define HM_OE(oe, statement) \
    do { \
        oe = statement; \
        if (oe != HM_OE_NONE) { \
            fprintf(stderr, "Error %s:%i\n", __FILE__, __LINE__); \
            goto LABEL_OUT; \
        }; \
    } while (0)

#define HM_OE_ERRNO_RESULT(oe, result, statement) \
    do { \
        result = statement; \
        if ((int64_t)result == -1) { \
            HM_OE(oe, errno); \
        }; \
    } while (0)
```

bootstrap/src/basis/ioring.c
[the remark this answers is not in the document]
cevans87: I haven't done this, but would we feel safe just changing the `unsigned` to `uint32_t` everywhere? I also really dislike the `unsigned`.

bootstrap/src/basis/ioring.c
```diff
@@ -0,0 +1,462 @@
...
+
+int io_uring_enter(unsigned int fd, uint32_t to_submit, uint32_t min_complete, uint32_t flags) {
+    return syscall(__NR_io_uring_enter, fd, to_submit, min_complete, flags, (size_t) NULL);
+}
+
+void hm_array_pp(int fd, int indent, unsigned len, unsigned *array) {
+    bool well_formed = true;
+    for (size_t i = 0; i < len && well_formed; i++) {
+        if (array[i] != i) {
+            well_formed = false;
```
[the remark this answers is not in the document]
cevans87: Opted for removing this fancy array formatting logic altogether.

bootstrap/src/basis/file.ml
jasone: Hmm, yeah, `=:<` is in `Rudiments`, but `iter` isn't. At one point I included all of `Range` in `Rudiments`, but that was a mess.

bootstrap/src/basis/ioring.c
```diff
@@ -0,0 +1,574 @@
...
+ *     actors managed by that executor may use the ioring instance, the instance is effectively used
+ *     in a single-threaded context and does not require any mutex protection.
+ *   - Actors emplace SQEs into the submission queue, but there is no need to actually submit (via
+ *     `io_uring_enter(2)`) until...
+ *     1. The submission queue is full.
+ *     2. The actor yields execution to the executor and there are SQEs sitting in the qeueue.
+ *     3. The actor blocks on an incomplete completion, possibly based on a SQE that is sitting in
+ *        the submission queue. In Hemlock, this will cause the actor to yield and is handled
+ *        identically to 2.
+ *   - We do not make any use of the submission queue indirection `array` data structure since as it
```
jasone, line 31:
```suggestion
 *   - We do not make any use of the submission queue indirection `array` data structure since it
```

bootstrap/src/basis/ioring.c
```diff
@@ -0,0 +1,574 @@
...
+ *     is intended to alleviate contention issues in multi-threaded use-cases. We initialize the
+ *     `array` once to point to their respective SQE in the SQE array.
```
jasone, line 33:
```suggestion
 *     `array` once to point to their respective SQEs in the SQE array.
```

bootstrap/src/basis/file.c
```diff
@@ -150,16 +154,29 @@ hm_basis_file_read_complete_inner(value a_bytes, value a_user_data) {
...
 
-    return caml_copy_int64((uint64_t)user_data);
+LABEL_OUT:
+    return submit_out(oe, user_data);
 }
 
-// val hm_basis_file_open_submit_inner: Basis.File.Flag.t -> uns -> Stdlib.Bytes.t >{os}->
-//   File.Open.t
+// hm_basis_file_open_submit_inner: Basis.File.Flag.t -> uns -> Stdlib.Bytes.t >{os}->
+//   (int * &File.Open.t)
```
jasone, line 179: FYI, our tuple type syntax is probably going to be `(int, &File.Open.t)`. IIRC we can pull this off (unlike OCaml) due to using postfix type paremetrization (like 1ML).

bootstrap/src/basis/ioring.c
```diff
@@ -265,86 +297,82 @@ hm_cqring_setup(void *vm, struct io_cqring_offsets const *offsets, hm_cqring_t *
...
-hm_opt_error_t
+void
 hm_ioring_setup(hm_ioring_t *ioring) {
     hm_opt_error_t oe = HM_OE_NONE;
 
     memset(ioring, 0, sizeof(*ioring));
-    HM_OE_ERRNO_RESULT(oe, ioring->fd, io_uring_setup(32, &ioring->params));
+    ioring->fd = io_uring_setup(HM_IORING_ENTRIES, &ioring->params);
+    if (ioring->fd == -1) {
+        abort();
```
jasone, line 317: We're going to have to recover from these when we integrate executors. I'd imagined this function returning an error, with the current implementation aborting in the caller. That makes later integration easier.

jasone (approved):

bootstrap/src/basis/ioring.c
cevans87: I can move the abort into `file.c` or `file.ml`. I think `file.ml` would be the more appropriate of the two.

PR #171, opened 2021-11-10
Add `.vscode/` to .gitignore

cevans87:


cevans87:
Editing from the web interface added some unwelcome commits. Fixing that.

jasone:
Shouldn't this just go in your personal config? As an example, I have `.*.swp` in my personal config to exclude vim swap files.

cevans87:
> Shouldn't this just go in your personal config?
Didn't realize that was a thing. Fixed.

PR #182, opened 2021-11-11
Attach CI to client push.

cevans87:
Associated `gh` extension PR: https://github.com/BranchTaken/gh-push/pull/1

### DIFF
--- a/.dockerfile
+++ b/.dockerfile
@@ -1,17 +1,13 @@
 ARG HEMLOCK_PLATFORM=$BUILDPLATFORM
 ARG HEMLOCK_UBUNTU_TAG
 FROM --platform=${HEMLOCK_PLATFORM} ubuntu:${HEMLOCK_UBUNTU_TAG} AS base
-RUN --mount=type=cache,target=/var/cache/apt \
-    --mount=type=cache,target=/var/lib/apt \
-    rm /etc/apt/apt.conf.d/docker-clean \
-    && echo 'Binary::apt::APT::Keep-Downloaded-Packages "true";' \
-        > /etc/apt/apt.conf.d/keep-cache \
-    && apt-get update \
+RUN apt-get update \
     && DEBIAN_FRONTEND=noninteractive apt-get install --no-install-recommends -y \
         ca-certificates \
         git \
         m4 \
         opam \
+        python3 \
         sudo \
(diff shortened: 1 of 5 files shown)


cevans87:
Note that I'm having to manually rerun the test every time I push. the `status` trigger I added should cause the test workflow to correctly trigger every time the commit status updates, but only after the workflow is in the `main` branch.

.dockerfile
```diff
@@ -41,14 +34,20 @@ RUN --mount=type=cache,target=/home/hemlock/.opam/download-cache,uid=1000,gid=10
...
     && rm Hemlock.opam
 
+FROM --platform=${HEMLOCK_PLATFORM} prod AS pre-push
+ARG HEMLOCK_PRE_PUSH_CLONE_PATH
+USER hemlock
+WORKDIR /home/hemlock/Hemlock
+COPY --chown=hemlock:hemlock ${HEMLOCK_PRE_PUSH_CLONE_PATH}/ ./
+RUN opam exec -- dune build \
+    && opam exec -- dune runtest \
+    && python3 .github/scripts/check_ocp_indent.py
```
jasone, line 44: At some point we'll want to refactor things, perhaps to run all scripts in the directory rather than having to manually invoke each one here.

jasone (approved):

.dockerfile
```diff
@@ -41,14 +34,22 @@ RUN --mount=type=cache,target=/home/hemlock/.opam/download-cache,uid=1000,gid=10
     && opam install -y --deps-only . \
     && rm Hemlock.opam
 
+FROM --platform=${HEMLOCK_PLATFORM} prod AS pre-push
+ARG HEMLOCK_PRE_PUSH_CLONE_PATH
+USER hemlock
+WORKDIR /home/hemlock/origin
+COPY --chown=hemlock:hemlock ${HEMLOCK_PRE_PUSH_CLONE_PATH}/ .
```
cevans87, line 41: The `gh-push` plugin wasn't properly setting this argument. Since it was an empty string, this line copied `/` to `.`. Since `/` is the base of our Hemlock repo, this set us up in a valid state for the `git clone` on line 43. The correct stuff got tested, but this was broken. I should have enforced that the arg be set.

PR #187, opened 2021-11-14
Revert "Attach CI to client push."

jasone:
This reverts commit 7860c215833b1a1acd3503192873d61631e47f1c.


jasone:
There are problems with this change that break CI in multiple ways, leave the local repo in a questionable state, etc. Reverting until it can be refined.

PR #194, opened 2021-11-15
Fixed push ci.

cevans87:
I left the commits separated for review in case you want to compare to the original implementation. I plan to squash before merging.

### DIFF
--- a/.dockerfile
+++ b/.dockerfile
@@ -1,17 +1,13 @@
 ARG HEMLOCK_PLATFORM=$BUILDPLATFORM
 ARG HEMLOCK_UBUNTU_TAG
 FROM --platform=${HEMLOCK_PLATFORM} ubuntu:${HEMLOCK_UBUNTU_TAG} AS base
-RUN --mount=type=cache,target=/var/cache/apt \
-    --mount=type=cache,target=/var/lib/apt \
-    rm /etc/apt/apt.conf.d/docker-clean \
-    && echo 'Binary::apt::APT::Keep-Downloaded-Packages "true";' \
-        > /etc/apt/apt.conf.d/keep-cache \
-    && apt-get update \
+RUN apt-get update \
     && DEBIAN_FRONTEND=noninteractive apt-get install --no-install-recommends -y \
         ca-certificates \
         git \
         m4 \
         opam \
+        python3 \
         sudo \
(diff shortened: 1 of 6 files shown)


.dockerfile
```diff
@@ -1,17 +1,13 @@
 ARG HEMLOCK_PLATFORM=$BUILDPLATFORM
 ARG HEMLOCK_UBUNTU_TAG
 FROM --platform=${HEMLOCK_PLATFORM} ubuntu:${HEMLOCK_UBUNTU_TAG} AS base
-RUN --mount=type=cache,target=/var/cache/apt \
-    --mount=type=cache,target=/var/lib/apt \
-    rm /etc/apt/apt.conf.d/docker-clean \
```
jasone, line 6: Can you educate me on why these mounts aren't needed anymore (here and below)?

jasone (approved):
This looks great!

The commit message refers to `origin/master` rather than `origin/main`.

Also, it looks like the commit message is wrapped at 100 columns. I've been wrapping at 80 columns ("best practice" is 72 for reasons related to email workflows), and I've started dropping the period from the first line to make it an intentional one-line synopsis sort of like single-line comments. TBH the 80-column wrapping is a pain... Oh, and I've been tempted to use Markdown in commit messages but haven't done so because even in the GitHub UI commit messages are rendered as plain text with the exception of links to issues (and perhaps other GitHub-specific linking), but even so, it's hard to clearly write commit messages without e.g. `literal` delimiters. Let's talk about how to format commit messages going forward.

.dockerfile
cevans87: We never actually needed them. They allowed us to keep the download artifacts for apt packages (similar story for the removed opam packages) in a docker-managed volume. Basically, for this single command, apt would have access to the same downloaded artifacts as previous runs and might be able to use them without actually downloading.

We really don't need this, since these stages rarely change. We're not re-downloading over and over like I'd worried we would. Also, while making changes to the dockerfile for this PR, I ran into a lot of problems with permissions and lock state in the opam caches. The best solution I think is to just remove the complexity.

cevans87: Rewrapped the commit message at 80 columns.

cevans87:
Seems to work well enough for now. Let's start using it.

PR #219, opened 2021-11-24
Complete slice refactor

### LINKED ISSUE #90: Complete slice refactor
`Slice` is now functorized, but only the `String`, `Bytes`, and `File` modules provide APIs which make operating on slices as ergonomic as on the corresponding containers. `Array.Slice` needs to be similarly refactored. Several other containers (`List`, `Ordset`, `Ordmap`) completely lack slices, even though they support cursors; these should all have slice support as well.

Tasks:

- [x] #218
  + [x] Convert `blit` to operate on slices rather than range-constrained arrays
  + [x] Split `?range` parameters out of `[pn]search` and provide equivalent functionality via `Slice.[pn]search`
- [x] Refactor `Bytes.Slice.pp` to use `Array.Slice.fmt`
- [x] ~~Refactor `List` to support slices~~
- [x] ~~#220~~
- [x] ~~Refactor `Ordmap` to support slices~~
- [x] ~~Refactor `Ordset` to support slices~~

jasone:
Will resolve #90.


cevans87 (approved):
LGTM

PR #239, opened 2022-12-07
Docker cleanup

cevans87:


docker-compose.yml
```diff
@@ -9,17 +9,9 @@ services:
...
-      - SYS_PTRACE
-    environment:
-      - SSH_AUTH_SOCK
     image: branchtaken/hemlock:prod
-    security_opt:
-      - seccomp:unconfined
-    privileged: true
     volumes:
       - .:/home/hemlock/Hemlock
-      - $SSH_AUTH_SOCK:$SSH_AUTH_SOCK
```
cevans87: [Here](https://github.com/cevans87/dotfiles/blob/a1d90017f5be040d5da5a87e1d30e836da908a83/hardlink_home/cevans/docker-compose.yml#L9-L27)'s an example of how I'm handling this.

docker-compose.yml
```diff
@@ -9,17 +9,7 @@ services:
...
-      - SYS_PTRACE
-    environment:
-      - SSH_AUTH_SOCK
     image: branchtaken/hemlock:prod
-    security_opt:
-      - seccomp:unconfined
-    privileged: true
-    volumes:
-      - .:/home/hemlock/Hemlock
-      - $SSH_AUTH_SOCK:$SSH_AUTH_SOCK
```
cevans87: [Here](https://github.com/cevans87/dotfiles/blob/a1d90017f5be040d5da5a87e1d30e836da908a83/hardlink_home/cevans/docker-compose.yml#L9-L27)'s an example of how I'm replicating this from a docker compose target I defined in my dotfiles.

Dockerfile
```diff
@@ -28,10 +27,7 @@ RUN opam init \
         --shell-setup \
         --yes \
     && opam switch create ${HEMLOCK_BOOTSTRAP_OCAML_VERSION} \
-    && opam install -y \
-        ocaml-lsp-server \
-        ocp-indent \
-        utop \
+    && opam install -y ocp-indent \
```
cevans87, line 27: [Here](https://github.com/cevans87/dotfiles/blob/master/opam/hemlock/general#L1-L2)'s an example of how I moved this into my dotfiles.

.env
```diff
@@ -1,5 +1,4 @@
 COMPOSE_DOCKER_CLI_BUILD=1
-COMPOSE_FILE=.docker-compose.yml
```
jasone, line 2: Do I understand correctly that the removed lines in this diff were redundant, in that they specified default values?

jasone (approved):
Re: `Remove .env file`, the commit message says, "1. We can now move the `pre-push` block [...]", but I don't see any corresponding changes to `gh-push`. Was this part of the commit message obsoleted by the final form of the changes, where `latest` automatically does everything required? FWIW, the `latest` name confused me, and if I understand the changes correctly, something like `tested`, `validated`, etc. would be closer to the image's intent. Case in point: As I refactored my custom wrappers, I initially switched from the `dev` target to the `latest` target, but `base` appears to be more appropriate.

Typo in `Remove .env file` commit message: `s/Then can now do this/They can now do this`

.env
cevans87: No. The filenames were changed from non-default values to default values.
- `.docker-compose.yml` -> `docker-compose.yml` (leading period removed)
- `.dockerfiled` -> `Dockerfile`

Dockerfile
```diff
@@ -29,6 +29,17 @@ RUN opam init \
...
+USER hemlock
+WORKDIR /home/hemlock/Hemlock
+COPY --chown=hemlock:hemlock /.git/ .git
+RUN git reset --hard \
+    && cd bootstrap \
+    && find . -type f -regex '.*\mli?' | xargs -- opam exec -- ocp-indent -i \
+    && git diff --exit-code \
+    && opam exec -- dune build \
+    && opam exec -- dune runtest
+
```
cevans87: Perhaps this would be more useful.

```
RUN git reset --hard

FROM --platform=${HEMLOCK_PLATFORM} latest AS tested
USER hemlock
WORKDIR /home/hemlock/Hemlock/bootstrap
RUN find . -type f -regex '.*\mli?' | xargs -- opam exec -- ocp-indent -i \
    && git diff --exit-code \
    && opam exec -- dune build \
    && opam exec -- dune runtest
```

cevans87: The idea being that `latest` includes a clean codebase. `tested` ensures it passes tests.

jasone: LGTM.

PR #243, opened 2024-06-28
Fixes io_uring permission error in Docker.

cevans87:
`io_uring_enter`, `io_uring_register`, and `io_uring_setup` were taken out of the Docker default seccomp profile (see
https://github.com/moby/moby/commit/891241e7e74d4aae6de5f6125574eb994f25e169). This caused our tests to fail with `permission denied` when calling any of those system calls during `docker compose build tested`.

This commit changes our seccomp profile to unconfined and runs our containers as privileged. Also, since images can only be built with the default profile, we can no longer run our tests during image build time. They are now run as a container command.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,11 +36,12 @@ WORKDIR /home/hemlock/Hemlock
 COPY --chown=hemlock:hemlock /.git/ .git
 RUN git reset --hard
 
-FROM --platform=${HEMLOCK_PLATFORM} latest AS tested
+FROM --platform=${HEMLOCK_PLATFORM} latest AS test
 USER hemlock
 WORKDIR /home/hemlock/Hemlock/bootstrap
-RUN find . -type f -regex '.*\mli?' | xargs -- opam exec -- ocp-indent -i \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -36,11 +36,12 @@ WORKDIR /home/hemlock/Hemlock
 COPY --chown=hemlock:hemlock /.git/ .git
 RUN git reset --hard
 
-FROM --platform=${HEMLOCK_PLATFORM} latest AS tested
+FROM --platform=${HEMLOCK_PLATFORM} latest AS test
 USER hemlock
 WORKDIR /home/hemlock/Hemlock/bootstrap
-RUN find . -type f -regex '.*\mli?' | xargs -- opam exec -- ocp-indent -i \
+CMD find . -type f -regex '.*\mli?' | xargs -- opam exec -- ocp-indent -i \
```
jasone, line 42: Hah, `s/RUN/CMD/` is the part I couldn't figure out from the docs!

jasone (approved):

PR #262, opened 2026-01-12
Glow-up pull-request C/I.

cevans87:
Three commits.

# 1. Fix ocp-indent errors.

Found by executing the following shell command in `Hemlock/bootstrap` directory.

```
find . -type f -regex '.*\mli?' \
        | grep -v -e '^\./test/hocc/' -e '^\./bin/hocc/Parse\.ml$' \
        | xargs -- opam exec -- ocp-indent -i \
     && git diff --exit-code
```

Before this diff, the command finished with an error. After, it finished
with success.

# 2.  Glow-up pull-request C/I.

This addresses the following.

1. Pull request CI currently breaks due to underlying issues with opam
  on the ubuntu:rolling image.

2. A laundry list of nice-to-haves that I wanted when initially building
  this CI are now well-supported.

  - Native ARM via the `ubuntu-24.04-arm` runs-on option.
  - Github Action runners now run bleeding-edge Linux kernel versions,
    so no workarounds are needed for our io_uring usage.
  - Docker provides a few excellent github actions steps
    (`docker/setup-buildx-action@v3', `docker/build-push-action@v5`)
    that make caching images in our GitHub Actions cache trivial.

3. User permission madness.

  - Docker did not deal well with uid:gid mismatches between the host
    and container. It now remaps the container root user to the
    the host (you) user.

    So, to get the container user to match us, we can do

    a. Zero work setting up our container user, use the container root
        user, and be happy.

    b. A lot of work setting up our container user by piping our own
        UID:GID through the image bake process, and be sad.

    > (you) is (container root), but (you) is not (container you)
    > unless (container you) is really (you).
    >
    > ~ docker
    
 # 3. Add `dev` docker service.
 
This reimplements the `Dockerfile` and `docker-compose.yml` (now called
`compose.yaml`, as it's the new, canonical name), hopefully making it
easier to use for development. There is only one service now. Mechanisms
for customization are documented in README.md and in compose.yml (which
is referred to by README.md).

I tested all of the suggestions documented in the compose.yaml except
SSH_AUTH_SOCK passthrough (I took the settings from a working
`docker-compose.yml` files I prototyped in my fork of Hemlock).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ bootstrap/_build/
 
 bootstrap/**/.merlin
 bootstrap/Hemlock.install
+**/compose.override.yaml
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,21 @@
+FROM ubuntu AS dev
+ARG HEMLOCK_EXTRA_APT_PACKAGES
+ARG HEMLOCK_EXTRA_OPAM_PACKAGES
+ARG HEMLOCK_GID
+ARG HEMLOCK_UID
+ARG HEMLOCK_OPAMSWITCH=4.14.2
+RUN --mount=type=bind,source=/bootstrap/Hemlock.opam,target=/Hemlock.opam \
+    userdel -r ubuntu \
+    && groupadd -g "${HEMLOCK_GID}" hemlock \
+    && useradd -d /home/hemlock -g "${HEMLOCK_GID}" -G sudo -m -u "${HEMLOCK_UID}" hemlock \
+    && apt-get update \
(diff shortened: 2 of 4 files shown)


cevans87:
@jasone This most likely breaks your local development workflow. Let me know how I can help unbreak it.

cevans87:
> I had add this in my script prior to the `docker compose` invocations in order to get my volumes to mount as `hemlock:hemlock` ownership rather than `root:root`:
> 
> ```shell
> export PODMAN_USERNS=keep-id:uid=1000,gid=1000
> ```
> 
> Maybe this should be part of `compose.yaml`?

I deleted it at some point to test the effect and forgot to add it back in. It's back in.

jasone (approved):

